PR #91, opened 2023-05-11
chore: add Postman collection

AikoPath:
Adding postman collection for easier development on the backend.

We should also add a more convenient way to get the auth token by using variables etc. in postman.


backend/Letters API.postman_collection.json
```diff
@@ -0,0 +1,144 @@
...
+			"response": []
+		},
+		{
+			"name": "POST templates",
+			"request": {
+				"method": "POST",
+				"header": [],
+				"body": {
+					"mode": "raw",
+					"raw": "{\n    \"html\": \"this is sample template\",\n    \"name\": \"sample template {{ placeholder1 }}\",\n    \"thumbnailS3Path\": \"s3path\",\n    \"fields\": \"[\\\"{{ placeholder1 }}\\\"]\"\n}",
```
gweiying, line 70: Just one small nit from me, @oversparkling has updated our validation for template creation on the `fields` array to validate that it contains an array of strings, instead of a stringified array. 

Would be nice to update this, the correct valid input would now be:
```
"fields": ["placeholder1"]
```

gweiying:
Looks great, I imported it and it worked like a breeze. Small suggestion on naming, can we remove the whitespace in the filename?

gweiying (approved):

gweiying (approved):

backend/Letters API.postman_collection.json
AikoPath: Thanks for the info. 👍
I've updated the request body accordingly.

PR #94, opened 2023-05-12
feat: validate bulk create

disKeith:
## Approach

- Updated `ValidationService` to `LettersValidationService`
- Updated `validateBulk` and `bulkCreate` to return 400 with
  - `message`: 'Number of rows exceeded max length of bulk create' | 'Malformed bulk create object'
  - `details`: array of invalid params (empty field | missing param | invalid params)
  - 400 Bad Request Example
```
"message": "Malformed bulk create object",
    "details": [
        {
            "id": 1,
            "param": "firstName",
            "message": "Field is empty"
        },
        {
            "id": 2,
            "param": "firstName",
            "message": "Missing param"
        },
        {
            "id": 3,
            "param": "gender",
            "message": "Invalid attribute in param"
        },
        {
            "id": 4,
            "param": "due",
            "message": "Invalid attribute in param"
        },
        {
            "id": 4,
            "param": "firstName",
            "message": "Missing param"
        },
        {
            "id": 4,
            "param": "lastName",
            "message": "Missing param"
        },
        {
            "id": 4,
            "param": "age",
            "message": "Missing param"
        },
        {
            "id": 4,
            "param": "city",
            "message": "Missing param"
        }
    ]
```


backend/src/letters/letters.service.ts
```diff
@@ -51,7 +57,21 @@ export class LettersService {
     const { templateId, letterParamMaps } = createBulkLetterDto
     const template = await this.templatesService.findOne(templateId)
     if (!template) throw new NotFoundException('Template not found')
-    // TODO: validation logic
+
+    const validationResult = this.lettersValidationService.validateBulk(
+      template.fields,
+      createBulkLetterDto.letterParamMaps,
```
gweiying, line 63: Small nit, since we've already unpacked the tuple in line 57, we can actually just use 
```const validationResult = this.lettersValidationService.validateBulk(
      template.fields,
      letterParamMaps)
```  here

File: backend/src/letters/letters-validation.service.ts [hunk 2]
```diff
@@ -1,29 +1,33 @@
...
 @Injectable()
-export class ValidationService {
-  bulkValidation(jsonStream: LetterParamMaps, fields: string) {
-    // Length validation
-    if (jsonStream.length >= BULK_MAX_ROW_LENGTH) {
-      throw new BadRequestException(
-        'Number of rows exceeded max length of bulk create',
-      )
+export class LettersValidationService {
+  validateBulk(template: string[], params: LetterParamMaps): ValidationResult {
```
gweiying, line 14: Suggestion, it might be nice to standardize the naming of the params here so that it matches the rest of the letter services. 

In this case, what might be more precise would be something like:
```
validateBulk(fields: string[], letterParamMaps: LetterParamMaps): ValidationResult {
```

backend/src/letters/letters-validation.service.ts
```diff
@@ -1,29 +1,33 @@
...
+        message: 'Number of rows exceeded max length of bulk create',
+      }
     }
 
-    // Field & attribution validation
-    const errorArray = []
-    for (let i = 0; i < jsonStream.length; i++) {
-      const obj = jsonStream[i]
+    for (let i = 0; i < params.length; i++) {
+      const obj = params[i]
```
gweiying, line 25: Another small suggestion on improving the naming of params here, `obj` is quite generic. If we rename `params` to `letterParamMaps`, then I would suggest `letterParamMap` here instead of `obj`. This would help us in the future in retracing through the code, as it's easier to understand what this `obj` actually is

backend/src/letters/letters.service.ts
```diff
@@ -51,7 +57,21 @@ export class LettersService {
...
+    )
+    if (!validationResult.success) {
+      const response = validationResult.errors
+        ? {
+            message: validationResult.message,
+            details: validationResult.errors,
+          }
+        : { message: validationResult.message }
+      throw new BadRequestException(response)
```
gweiying, line 72: Just a small suggestion (totally non-blocking), but I think for brevity's sake it might be ok just to return the validation result in the BadRequestException, e.g. `throw new BadRequestException(validationResult)`.

backend/src/letters/letters-validation.service.ts
```diff
@@ -32,7 +36,7 @@ export class ValidationService {
           })
         }
       }
-      for (const attr of fields) {
+      for (const attr of template) {
```
gweiying, line 39: Again, small suggestion here for continuity, but here we could name this `for (const field of fields)` which might make it easier to read once again.

backend/src/letters/letters-validation.service.ts
```diff
@@ -32,7 +36,7 @@ export class ValidationService {
...
         // If object does not have the attribute in the template
         if (!Object.prototype.hasOwnProperty.call(obj, attr)) {
           errorArray.push({
```
gweiying, line 42: Just a small suggestion, I think it'd be ok to return the same error for both of these errors, since both are around missing params!

gweiying (approved):
Thanks for the PR! I left some comments. Overall the logic looks good and I think the PR is basically good to go, but I had some small suggestions on naming as I felt that some of the variable naming in your application was either too generic (difficult to follow the logic), or out of odds with the convention in the rest of the application. 

Let me know what you think of the suggestions, or if you have other ideas!

disKeith:
thanks @gweiying, the suggestions to improve readability and brevity was definitely better! also, the intent of previously separating the messages for empty and missing fields was to have greater resolution of the invalid field(s) - but if it's not needed then will collapse them together as suggested!  

changes in reflected in https://github.com/opengovsg/letters/pull/94/commits/572204f5b15be4c20f87d1b9e683ff485a9064cd

backend/src/letters/letters-validation.service.ts
```diff
@@ -11,23 +11,26 @@ class ValidationResult {
 
 @Injectable()
 export class LettersValidationService {
-  validateBulk(template: string[], params: LetterParamMaps): ValidationResult {
+  validateBulk(
+    fields: string[],
+    LetterParamMaps: LetterParamMaps,
```
gweiying, line 16: nit, small first letter for variables! only types tend to have capitalized first letters

gweiying (approved):
Thanks for the quick turnaround! One last comment on the variable naming, thanks!!

backend/src/letters/letters-validation.service.ts
disKeith: thanks for catching this, resolved in https://github.com/opengovsg/letters/pull/94/commits/996c857b7ac4c6fa1c99c3056690ecabfce17c5c

PR #95, opened 2023-05-12
Feat/sample csv bulk

oversparkling:
## Context

Closes [this](https://www.notion.so/opengov/Get-sample-CSV-for-issuing-bulk-letter-ff20562e2422495a921b032a0620f158?pvs=4)

## Approach

Create a util function that handles 2 main flows:
1. Taking in the fields of the templates, and generate a sample csv for the user to download
2. Having the response of the bulk creation, convert the json object into csv for the user

**Improvements**:

Currently the 2 fields of the json to csv converter are `publicId` and `createdAt` only. We should aim to accept more fields in the long run to make this function scalable


frontend/src/features/CsvParser.ts
```diff
@@ -0,0 +1,21 @@
...
+    ...bulkGeneratedResponse.map((item) => [item.publicId, item.createdAt]),
+  ]
+}
+
+export const convertFieldToCsv = (csvStringArray?: string[]) => {
+  if (csvStringArray) {
+    const blob = new Blob([csvStringArray.join(',')], {
+      type: 'text/csv;charset=utf-8',
+    })
+    saveAs(blob, 'Discharge Memo.csv')
```
gweiying, line 19: Nit, the name here seems to be hardcoded, can we update this to either:
a) not hardcoded (can take from the Template Name)
b) Generic name such as "Sample Template.csv"

I'm ambivalent about either, please help check with Pranu!

frontend/src/features/templates/components/SampleCsvCard.tsx
```diff
@@ -1,18 +1,31 @@
...
-  </ToggleCard>
-)
+}: SampleCsvCardProps) => {
+  return (
+    <ToggleCard {...toggleCardProps}>
+      <CardHeader>Download the sample .CSV file</CardHeader>
+      <Button
+        onClick={() => {
+          //TODO: Error handling if templateFields are null or templates
+          convertFieldToCsv(toggleCardProps.templateFields)
```
gweiying, line 23: Nit, can we unpack `templateFields` as a separate variable on SampleCsvCard instead of from `toggleCardProps`, since it's not really a type of `toggleCardProps`?

To be more precise, we can do something like this:
```
export const SampleCsvCard = ({
  onCompletion,
  templateFields,
  ...toggleCardProps
}: SampleCsvCardProps) 
```

frontend/src/features/templates/components/SampleCsvCard.tsx
```diff
@@ -1,18 +1,31 @@
...
-    <Button onClick={onCompletion}>Download</Button>
-  </ToggleCard>
-)
+}: SampleCsvCardProps) => {
+  return (
+    <ToggleCard {...toggleCardProps}>
+      <CardHeader>Download the sample .CSV file</CardHeader>
+      <Button
+        onClick={() => {
+          //TODO: Error handling if templateFields are null or templates
```
gweiying, line 22: Instead of leaving this as a TODO, can we add a simple fix to disable the button if `templateFields` is undefined?

frontend/src/features/CsvParser.ts
```diff
@@ -0,0 +1,21 @@
+import saveAs from 'file-saver'
```
gweiying, line 1: nit on code organization, these functions are largely util-related, can we move this into `utils/csv-parser.ts`

frontend/src/features/CsvParser.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import { GetLetterPublicDto } from '~shared/dtos/get-letter.dto'
+
+export const convertJsonToCsvString = (
```
gweiying, line 5: This seems like an unused function for now, can help me clarify what this is doing?

gweiying (approved):
The changes look largely good, just left some inline clarifications on code organization and suggested variable naming!

frontend/src/features/CsvParser.ts
oversparkling: So the idea is that there is 2 flow for this csv parser.

1. Downloading the sample csv: In this case we are just using the `convertFieldToCsv` which essentially takes the template fields (`["name","location","time"]`) and convert it to a csv

2. Downloading the response from bulk creation: our input for this would be the json response of the call (`GetPublicLetterDTO[]`) and what `convertJsonToCsvString` does is converting this json into a csv string form that will then be piped into the `convertFieldToCsv` function 

Not sure if the naming is good though 🤔

frontend/src/utils/CsvParser.ts
```diff
@@ -0,0 +1,25 @@
...
+import { GetLetterPublicDto } from '~shared/dtos/get-letter.dto'
+
+export const convertJsonToCsvString = (
+  bulkGeneratedResponse: GetLetterPublicDto[],
+) => {
+  return [
+    ['publicId', 'createdAt'],
+    ...bulkGeneratedResponse.map((item) => [item.publicId, item.createdAt]),
+  ]
+}
```
gweiying, line 12: Minor suggestion on util functions, would be cleaner if this was a generic templating function, and the required fields just be variables that can be specified.

gweiying (approved):
looks good, one last suggestion on the utility function! Would be nice if it was reusable and not specific to this one usecase, especially since the name is quite generic.

frontend/src/utils/CsvParser.ts
gweiying: Sorry, I think I was not quite precise enough about what I meant by generic templating function, I was thinking about something like the following (provided the method types below):

```
export const convertJsonArrToCsvString = (
  jsonArr: { [key: string]: string }[],
  selectedFields: string[],
) => {...// code logic here}
```

PR #97, opened 2023-05-15
fix: change class validation

oversparkling:
## Context

Currently requests made to creation of template have to use arrays in a weird way (storing the actual array as a string). Ideally we want to use the native array type for this instead of storing it as a literal string

Change the class-validation

## Before & After Screenshots

**BEFORE**:
Before change
Screen Shot 2023-05-15 at 11 08 16 AM

**AFTER**:
After change
Screen Shot 2023-05-15 at 11 08 44 AM


gweiying (approved):
lgtm!

PR #98, opened 2023-05-15
feat: add issued letters page

halfwhole:
## Context

We want to add the "Issued Letters" page

## Approach

Added a table that shows all issued letters to the issued letters page.

The table applies pagination with limit and offset. (There may be performance concerns with these paginated queries, but not so concerned with it right now!)

**Improvements**:

- Renamed `DashboardPage` to `TemplatesPage` and `DashboardBody` to `TemplatesBody` (this helps with disambiguation, because the dashboard comprises both the templates page and the issued letters page)

Some TODOs to do in the future:
- Implement the "Shared with" column. Right now, it doesn't seem that a letter record in the DB stores information on who the letter was sent to, so we might need to make a new migration for that
- Implement sorting by different columns. Right now, it defaults to reverse chronological order. Doing this doesn't seem that important, so skipping it for now (would this be okay?)
- ~Properly retrieve the host name for the URL (i.e. `letters.[beta.].gov.sg`) through an env variable or something~ done through `document.location.host`

## Screenshots

Screenshot 2023-05-15 at 5 13 48 PM

## Tests

Saw we don't have really unit tests, so skipping it for now


gweiying (approved):
looks great to me! can't find anything that I would change at all, brilliant stuff

PR #102, opened 2023-05-16
chore: add datadog rum

halfwhole:
## Context

We want to add Datadog RUM to track user behaviour metrics

Closes [Notion task](https://www.notion.so/opengov/Set-up-Datadog-RUM-so-that-we-have-user-behavior-metrics-c0045613c1ab4fe68714bf735a4a222a?pvs=4)

## Approach

Followed [Datadog's RUM setup instructions](https://docs.datadoghq.com/real_user_monitoring/browser/) and [Redeem's Datadog sharing slides](https://docs.google.com/presentation/d/1s1Mho2krp46sdGUNbg8q7-zRDWm_SC-NbIoSSiNfVOQ/edit#slide=id.g11b1a1b1dce_0_111). Also edited the content security policy according to [Datadog's guidelines](https://docs.datadoghq.com/real_user_monitoring/faq/content_security_policy/).

Settings:
- 100% sample and replay rate, i.e. all user sessions will be tracked with replay functionality
- User inputs are masked

In the future, we'd probably want to find a way to enable datadog RUM for admin routes only, while excluding the public routes, because:
- We don't want to impose greater data or bandwidth burdens on MOPs
- We expect lots of letter visits from MOPs, and it would be expensive and wasteful and not very helpful to track them

## Tests

- [x] Tested locally and verified that RUM sessions are collected on Datadog
- [x] Tested on staging and verified that RUM sessions are collected on Datadog

## Deploy Notes

**New dependencies**:

- `@datadog/browser-rum` : for datadog RUM


backend/src/middlewares/helmet.middleware.ts
```diff
@@ -15,8 +15,8 @@ export class HelmetMiddleware implements NestMiddleware {
           defaultSrc: ["'self'"],
           baseUri: ["'self'"],
           blockAllMixedContent: [],
-          connectSrc: ["'self'"],
-          workerSrc: [],
+          connectSrc: ["'self'", 'https://*.browser-intake-datadoghq.com'],
+          workerSrc: ['blob:'],
```
gweiying, line 19: Just adding a small inline note here that allowing `workerSrc: blob:` in our CSP isn't ideal practice but seems to be the only workaround for now, let's monitor this issue and fix when possible. 
https://github.com/DataDog/browser-sdk/issues/1578

gweiying (approved):
lgtm! thanks for setting up

PR #104, opened 2023-05-17
chore: add uat ci to deploy app for load test

AikoPath:
Create UAT environment for load testing


AikoPath:
This PR is intended to stay un-merged as we don't need a permanent UAT environment but want to be able to provision this at any time for further load/performance tests.

PR #106, opened 2023-05-18
Load test to verify that the MVP is able to support the first use case 

AikoPath:
This load test was executed on Thu 18 May.

Results and details can be found in Notion.
https://www.notion.so/opengov/Load-Test-9a9289f78b8f4a589a03f913fd67a36f?pvs=4


gweiying (approved):
lgtm! thanks for documenting the notion findings too!

PR #107, opened 2023-05-18
Fix/copy

gweiying:
## Context

This PR updates the copy and assets throughout the frontend to add more information about Letters. 

Changes:
- LandingPage copy and assets
- Error page assets
- Header logos
- Website description and favicons

Additional logic change was added on the public letters retrieval hook to disable retries, so that the pubilc user will be redirected to the Error page when accessing a non-existent letter upon receiving a 404, instead of retrying thrice. 

## Before & After Screenshots
**AFTER**:
localhost_3000_
localhost_3000_letters_8QEkTAM2zqQ9WzAR1
localhost_3000_admin_templates


AikoPath:
Found one instance of "Highway" on the landing page, just pushed a commit to change that to "Letters"

AikoPath (approved):
LGMT

PR #108, opened 2023-05-18
Fix/dd tracer

gweiying:
## Context

Implements changes in https://github.com/opengovsg/ts-template/pull/2872. Quoting directly from the PR:
> Our traces currently don't have accompanied logs due to improper initialisation order as described in https://github.com/DataDog/dd-trace-js/issues/2189#issuecomment-1188455882. This PR moves importing of tracer at the start (before pino is instantiated/imported)

Thanks @sheikhshack for the fix and @halfwhole for noticing it.

Tested on staging and works now, see screenshots below of our unhandled Postman email error, before and after fixing the trace order.

## Screenshots
**BEFORE**
Screenshot 2023-05-19 at 9 25 07 AM
**AFTER**
Screenshot 2023-05-19 at 9 25 00 AM

### DIFF
--- a/backend/src/app.module.ts
+++ b/backend/src/app.module.ts
@@ -1,6 +1,3 @@
-// Initialize Datadog tracer
-import 'tracing'
-
 import { MiddlewareConsumer, Module, NestModule } from '@nestjs/common'
 import { APP_PIPE } from '@nestjs/core'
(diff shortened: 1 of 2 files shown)


halfwhole (approved):
lgtm thanks for the fix!

PR #110, opened 2023-05-19
Release 1.1

gweiying:


CHANGELOG.md
```diff
@@ -0,0 +1,44 @@
...
+
+Generated by [`auto-changelog`](https://github.com/CookPete/auto-changelog).
+
+#### 1.1.0
+
+- Chore/release versioning [`#109`](https://github.com/opengovsg/letters/pull/109)
+- Fix/copy [`#107`](https://github.com/opengovsg/letters/pull/107)
+- fix: dd tracer import location [`#108`](https://github.com/opengovsg/letters/pull/108)
+- chore: add otp secret to ecs-task-definition [`#105`](https://github.com/opengovsg/letters/pull/105)
+- Load test to verify that the MVP is able to support the first use case  [`#106`](https://github.com/opengovsg/letters/pull/106)
```
halfwhole, line 13: what happened to the commits under v1.0.0 though :o are we beginning at v1.1.0?

halfwhole (approved):
🙌

CHANGELOG.md
gweiying: good catch, I'm actually not sure why this happened... lemme double check

gweiying: well it's abit weird, since we didn't make any commits between 1.0.0 and 1.1.0, now our initial release will have 2 version-tags. do you have any good ideas for resolving this? @halfwhole

halfwhole: perhaps we could set and push the tag `v1.0.0` at where the `production` branch currently is? if we're okay with setting that one to be 1.0.0 and the current one to be 1.1.0?

PR #112, opened 2023-05-19
feat: sanitize html for letters and templates

halfwhole:
## Context

For security, we want to ensure that the letters HTML that's displayed to the public is safe

Closes this [Notion task](https://www.notion.so/opengov/Eng-HTML-sanitisation-18e47a199c1e46eca9d414d02a79a158?pvs=4) (more details are available on that page)

(For an example of using DOMPurify, see Isomer's sanitiser at https://github.com/isomerpages/isomercms-backend/blob/5b0a50fab58255f41c0cf5b5651936cf91bdf7a7/src/services/utilServices/Sanitizer.ts)

## Approach

Use [DOMPurify](https://github.com/cure53/DOMPurify) to sanitise HTML. Although [TinyMCE 6 already comes built-in with DOMPurify](https://www.tiny.cloud/docs/tinymce/6/security/#sanitizing-html-input-to-protect-against-xss-attacks), we should also sanitise HTML on the server-side and client-side (when displaying using `dangerouslySetInnerHtml`).

We sanitise the HTML in three places:
1. [Backend] On template creation
2. [Backend] On letter creation: this is needed even if the template has already been sanitized, because the field inputs from users might potentially contain unsafe content
3. [Frontend] On letter display

Configurations
- Disallow SVG and MathML
  - TinyMCE considers SVG to be a potential source of XSS attacks, and so they disable it by default. It should be safe to allow SVGs since we're sanitising it with DOMPurify anyway, but still we might want to err on the safe side unless we really need SVGs
  - TinyMCE also appears to block MathML by default. MathML is kinda niche and may cause security issues, so again unless we really need it, it's probably better to just disallow it
- If we want to, we can further restrict the allowed HTML elements and attributes?

In the future, these configurations on DOMPurify should be made to match those on TinyMCE (when we add the template HTML editor later on)

## Tests

- [x] Manual tests: ensure removal of svg, mathml, and unsafe HTML elements and attributes (e.g. `<script>`, `onload`)
- [x] Unit tests: same as above

## Deploy Notes

**New dependencies**:

- `isomorphic-dompurify` : simple wrapper around `dompurify` and `jsdom`, it allows both the frontend and backend to use DOMPurify the same way (otherwise [running it server-side vs. client-side is different](https://github.com/cure53/DOMPurify#running-dompurify-on-the-server))

## Risks

Overall it seems pretty safe to use DOMPurify, and using it would likely guard against most possible threats. DOMPurify appears to be a well-maintained, popular and secure library that's helmed by Cure53, a reputable security organisation who seems like they know what they're doing. For more info on DOMPurify's security goals and what it protects and doesn't protect against, see their page on [Security Goals & Threat Model](https://github.com/cure53/DOMPurify/wiki/Security-Goals-&-Threat-Model)

Also, only public officers or agencies will be allowed to create templates (in the future), so the risk of these templates and letters containing malicious code seems rather low.

We may also want to look into CSPs as another security measure on top of HTML sanitisation


backend/src/letters/letters.service.ts
```diff
@@ -32,22 +33,27 @@ export class LettersService {
...
+      this.sanitizeLetter(createLetterDto),
+    )
+    const letters = this.repository.create(sanitizedCreateLetterDtos)
+    return entityManager.save(letters)
+  }
+
+  private sanitizeLetter(createLetterDto: CreateLetterDto): CreateLetterDto {
+    return {
+      ...createLetterDto,
+      issuedLetter: sanitizeHtml(createLetterDto.issuedLetter),
```
gweiying, line 55: What do you think of abstracting this function (and a bulk version of this) out to a separate service, similar to `letters-rendering` and `letters-validation`? Off the top of my head, if we want to add additional logic for sanitization such as returning a custom error or something, it might be nice to separate this from the rest of the code.

What do you think?

backend/src/letters/letters.service.ts
```diff
@@ -32,22 +33,27 @@ export class LettersService {
...
     entityManager: EntityManager,
-  ): Promise<Letter | Letter[]> {
-    let created
-    if (Array.isArray(toCreate)) {
-      created = this.repository.create(toCreate as Partial<Letter>[])
-      console.log(created)
-    } else {
-      created = this.repository.create(toCreate as Partial<Letter>)
+  ): Promise<Letter[]> {
+    const sanitizedCreateLetterDtos = createLetterDtos.map((createLetterDto) =>
```
gweiying, line 45: Curious, why did you choose to sanitize inside the service method `createWithTransaction` as opposed to `bulkCreate` where the other services such as rendering and validation live?

gweiying:
Let's have a short discussion offline, I had some questions about code organization, was hoping to hear your thoughts!

backend/src/letters/letters.service.ts
halfwhole: was thinking to put this as close to the actual logic of letter creation as possible, would it make sense here?

backend/src/letters/letters.service.ts
halfwhole: sounds good! have created a separate service `letters-sanitization` for this, to deal with letter DTO sanitization 
https://github.com/opengovsg/letters/pull/112/commits/659b1243f27dceea8955df8667f198b40f745058

gweiying (approved):
lgtm! thanks for the discussion, I can see benefits to both ways of code organization so I think fine to leave it as is for now!

PR #118, opened 2023-05-22
chore: fix max bulk creation length check

gweiying:
## Context
Fixes check for `BULK_MAX_ROW_LENGTH`

If our `BULK_MAX_ROW_LENGTH` is 500, we should allow for the creation of 500 letters, not 499, so we should use the `>` check not `>=`.

## Tests
Tested on staging, load-wise we are able to handle creation of 500 links. Max CPU utilization in cluster went from 5% to 9%,  but average utilization in cluster remained stable around 2%. No notable spikes in DB usage.


halfwhole (approved):
lgtm!

PR #119, opened 2023-05-22
fix: error message more user friendly when the otp call to postman fails

AikoPath:
Provide a more helpful error message to the user in case the call to postman fails.
Reusing the error message from go.gov.sg

**Before:**
Screenshot 2023-05-22 at 6 20 45 PM

**After:**
Screenshot 2023-05-22 at 6 40 56 PM


gweiying (approved):
lgtm!

gweiying (approved):
Sorry, just noticed. Is there a particular reason why we might want to throw 500 here instead of a 400 error, since the application has handled this error already?

PR #120, opened 2023-05-23
refactor: types for bulk letter validation errors

halfwhole:
## Context

We want to implement the error flow, where users can see the validation errors in the CSV they uploaded

But the validation error types currently reside only in the backend, and so it's not usable by the frontend

Part of [Notion task](https://www.notion.so/opengov/Eng-bulk-creation-error-handling-31e044846ff049f09ed4cfc2f4db0b8a?pvs=4)

## Approach

Define the validation error types and move them to `/shared`, so that it can be used by the frontend in the future


gweiying (approved):
great idea, thanks for suggesting this

PR #123, opened 2023-05-24
feat: show upload csv errors on bulk issue drawer

halfwhole:
## Context

We want to show useful errors to the user when there are issues with their uploaded CSV (when issuing bulk letters)

Closes [Notion task](https://www.notion.so/opengov/Eng-bulk-creation-error-handling-31e044846ff049f09ed4cfc2f4db0b8a?pvs=4)

## Approach

Show "N errors detected" if there are errors with the uploaded CSV, and show a table containing all the rows in the uploaded CSV with errors.

Added two new frontend states to the drawer:
- `uploadCsvErrors`: stores the response errors, which corresponds to errors in the uploaded CSV
- `isShowUploadCsvErrors`: boolean that controls whether we display the main form view, or the upload csv error view (table view)
  - this boolean approach might not be the best, we'll need to refactor it once we add in the final step of downloading the generated csv

**Bugs and fixes**:
- Fixed a minor navigation bug in `onClose()`
- Discovered a minor issue: if the user provides more params than requested in the uploaded CSV, it'll be grouped and sent to the backend under `__parsed_extra` ([papaparse docs](https://www.papaparse.com/docs))

TODOs:
- Group all the errors for the same CSV row into the same table row, instead of splitting it into multiple table rows
  - This might be best implemented by making not just a frontend change, but changes to the DTO and backend response as well?
  - Hope to address this in another PR, otherwise adding it here is fine too!

## Screenshots & Videos

Screenshot 2023-05-24 at 6 01 58 PM
Screenshot 2023-05-24 at 6 02 20 PM

https://github.com/opengovsg/letters/assets/41856541/120f572a-ed28-4015-b9a7-83bb6225889a

## Tests

- E2E tests, we should have this someday :(

### DIFF
--- a/frontend/src/features/issue/BulkIssueDrawer.tsx
+++ b/frontend/src/features/issue/BulkIssueDrawer.tsx
@@ -1,5 +1,6 @@
 import {
   Button,
+  Card,
   Drawer,
   DrawerBody,
   DrawerCloseButton,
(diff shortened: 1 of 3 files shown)


AikoPath (approved):
> // TODO: group all errors of the same row number together. should this be done on the frontend or backend?

I think this can be done cleanly on the backend.
I am thinking of a structured response, something like this:

```
...
{ 
 2: [ list of errors of row 2],
 4: [ list of errors of row 4]
}
```

Additionally, I am wondering if we should also group errors by column in case every row has the same error. E.g. if a user re-uses and old template where a complete column is missing.
Instead of reporting 100 rows with the error "field3 is missing" we could just summarize this to one error.

What do you think?


All this being said, I think that can be done in a different PR and this PR can get merged to unblock the next stroy (which might change a few things in this file anyway).

halfwhole:
> Additionally, I am wondering if we should also group errors by column in case every row has the same error. E.g. if a user re-uses and old template where a complete column is missing.
> Instead of reporting 100 rows with the error "field3 is missing" we could just summarize this to one error.

yep, sounds like a good idea!

PR #124, opened 2023-05-24
Return letters data to user after bulk creation

AikoPath:
This response will be used in the CSV that can be downloaded by the user after bulk creation.

### DIFF
--- a/shared/src/dtos/letters.dto.ts
+++ b/shared/src/dtos/letters.dto.ts
@@ -56,7 +56,7 @@ export class GetLettersDto {
 }
 
 export class GetBulkLettersDto {
-    shortLink: string;
-    createdAt: Date;
-    [key: string]: any;
+  shortLink: string
+  createdAt: Date;
+  [key: string]: any
 }
--- a/shared/src/dtos/letters.dto.ts
+++ b/shared/src/dtos/letters.dto.ts
@@ -58,5 +58,5 @@ export class GetLettersDto {
 export class GetBulkLettersDto {
   shortLink: string
   createdAt: Date;
-  [key: string]: any
+  [key: string]: string | Date; // using `| Date` to allow TS to have a Date field sibling to the dynamic ones
 }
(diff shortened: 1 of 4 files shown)


shared/src/dtos/letters.dto.ts
```diff
@@ -54,3 +54,9 @@ export class GetLettersDto {
   letters: GetLetterDto[]
   count: number
 }
+
+export class GetBulkLettersDto {
+  shortLink: string
+  createdAt: Date;
+  [key: string]: any
```
halfwhole, line 61: might be missing some context for this, but it seems that `shortLink` is currently hardcoded to an empty string, and for MVP at least we won't be needing it since go.gov.sg shortlinks aren't in scope. could it be helpful to return `publicId` instead, which seems sufficient to construct the link `letters.[beta.]gov.sg/letters/:publicId` on the frontend csv?

also, for a bit more cleanliness, would it be nice to separate the extra parameters from `letterParamMaps` into a separate field, such as `params` or `paramMap`?

```
  ...
  params: {
    [key: string]: string
  }
```

halfwhole (approved):
lgtm! just have some questions about the response DTO format:

shared/src/dtos/letters.dto.ts
AikoPath: Good point!

I've been considering of `publicId` over `shortLink` and went with `shortLink` in hopes that we don't need to change this again this once we have the shortlink.
I was wondering if we could just use the actual letters link in the mean time as shortlink.
That way we don't have to change the code later AND the user already gets a working link right now without manually constructing one.
But I'm happy to go with either solution.
Checking with the team on whether an actual letters link as shortlink makes sense for the product.

Regarding your point on `letterParamMaps`: this whole response is only used to be mapped directly into a flat csv file that can be downloaded by the user. I think having the DTO match the structure that we use on this data for makes this straight forward.

PR #127, opened 2023-05-25
feat: add legal copy to letter

halfwhole:
## Context

We want to add the letter ID and legal copy below the letter

Closes [Notion task](https://www.notion.so/opengov/Eng-frontend-letters-UI-view-6114345a704b4ccab9445d89215b2ca0)

## Approach

Add an extra footer below the letter that contains the letter ID and legal copy.

Some choices made, in the process of confirming:
- This footer is *not* included as a part of the main letter, and hence will *not* be included in the downloaded PDF
- ~If the letter has too small a width, the footer will be wider than the letter (see screenshot 2). This is because we might not want the footer to inflate the actual dimensions of the letter HTML~ (see below)

Updated 12 Jun:
- Confirmed that footer is not included as part of downloaded PDF as the effort is too high and we're not sure if we want to keep the download PDF feature
- The letter width is now fixed, so it the footer will be the same width as the letter.

## Screenshots

Screenshot 2023-05-25 at 5 55 07 PM

localhost_3000_letters_ql6SXbGIOTuIGlBH


AikoPath (approved):
LGTM

PR #128, opened 2023-05-26
Let admin user download a csv file with the details after bulk generation of letters 

AikoPath:
## Context

When an admin successfully uploads a csv file to bulk generate letters this PR enables them to download a csv with the details (parameters, short link, ...) of the generated letters as a csv file.

Closes [Notion](https://www.notion.so/opengov/Eng-bulk-creation-download-frontend-a24a42b0937b44779b197810da0467d6?pvs=4)

## Approach

On successful response of the call to `/api/letters/bulks`, two state variables are populated with the letters details and a success respectively.
The drawer body conditionally renders the screen to download the CSV depending on this success state and lets the user download the csv which gets generated by passing the letters details to `jsonArrToCsv()` from `~utils/csvUtils`.

**Improvements**:

- Pluralizes screen content  like letter**s** and error**s** if needed.
- Updates GetBulkLettersDto to match expectation of the csv file

## Before & After Screenshots

**BEFORE**:
before csv download

**AFTER**:
after csv download

## Risks

As multiple parties have been working on the sames 
E.g. 
- the toast showing the number of generated letters is not in focus anymore.
- the styling of the Drawer and its child component does not exactly match the Figma design.
- - It would be good to sync up offline with everyone to discuss the final pending changes.


frontend/src/features/issue/BulkIssueDrawer.tsx
```diff
@@ -50,16 +55,21 @@ export const BulkIssueDrawer = (): JSX.Element => {
   const [uploadCsvErrors, setUploadCsvErrors] = useState<
     BulkLetterValidationResultError[]
   >([])
+  const [isUploadSuccess, setIsUploadSuccess] = useState(false)
```
gweiying, line 58: suggestion, instead of creating a new variable to store success state, you can consider using the `isSuccess` variable returned by the [mutation](https://tanstack.com/query/latest/docs/react/reference/useMutation), for example

```
const { mutateAsync, isLoading, isSuccess: isUploadSuccess } = useCreateBulkLetterMutation({
    onSuccess: (res) => {...}
```

frontend/src/features/issue/BulkIssueDrawer.tsx
```diff
@@ -50,16 +55,21 @@ export const BulkIssueDrawer = (): JSX.Element => {
...
+  const [bulkLetters, setBulkLetters] = useState<GetBulkLettersDto[]>([])
```
gweiying, line 59: similar suggestion, instead of creating a new variable to store data, you can consider using the `data` variable returned by the [mutation](https://tanstack.com/query/latest/docs/react/reference/useMutation), for example
const { mutateAsync, isLoading, data: bulkLetters } = useCreateBulkLetterMutation({
    onSuccess: (res) => {...}
    ...
})

frontend/src/features/issue/BulkIssueDrawer.tsx
```diff
@@ -50,16 +55,21 @@ export const BulkIssueDrawer = (): JSX.Element => {
...
   const navigate = useNavigate()
   const toast = useToast()
 
+  const pluraliseIfNeeded = (collection: any[], singular: string): string =>
```
gweiying, line 63: This is a good idea! Can we move it into the utils folder so we can use it throughout the frontend?

gweiying (approved):
overall looks good, just a couple of suggestions on using the state management variables provided out-of-the-box by react query. would help to reduce the number of variables we need to manage manually!

regarding the risks mentioned above, think it's a good idea to discuss these things but can be non-blocking for the PR!

frontend/src/features/issue/BulkIssueDrawer.tsx
AikoPath: Good point!
And thanks for sharing the useMutation docs.

frontend/src/features/issue/BulkIssueDrawer.tsx
AikoPath: While this is an equally good idea as the `isSuccess` above, in practices this is a little bit more tricky as the `data` field is of type `GetBulkLettersDto[] | undefined` (since the success can't be guaranteed for every call).
Which adds extra code(casting/guard clause) to handle this.

In this case I find a state of type `GetBulkLettersDto[]` a little bit more maintainable.

frontend/src/features/issue/BulkIssueDrawer.tsx
AikoPath: Done.
👍

PR #132, opened 2023-05-29
Serve fallback image for template preview

AikoPath:
## Context

In case the preview image of the template can not be loaded a fallback image gets shown instead.

Closes [Default image for letter template](https://www.notion.so/opengov/Design-Default-image-for-letter-template-766e2ac94dd64058994e6eabb9ebe799?pvs=4)

## Approach

Using `fallbackSrc` on `Chakra-ui`s `Image`.

## Improvments

In order to achieve the figma design for the fallback image, it was also necessary to update the general card design to match.

## Before & After Screenshots

**BEFORE**:
Screenshot 2023-05-29 at 1 41 29 PM

**AFTER**:
Screenshot 2023-05-29 at 9 58 55 PM


gweiying (approved):
lgtm!

PR #133, opened 2023-05-29
Return letter link in csv after bulk creation

AikoPath:
## Context

Previously no link was returned to the user.

## Approach

The bulk creation response now contains the public id of the letters.
This is used to construct the letters link on the front end before including it in the csv that the admin can download.

## Improvements:
- Rename `GetBulkLettersDto` to `GetBulkLetterDto` as that is the correct naming. Thanks Alexis for calling this out.
- Change `createdAt` and `shortlink` in the  CSV to be "Date of Issue" and "Letter Link" to be more user friendly.

## Before & After Screenshots

**BEFORE**:
before link included

**AFTER**:
Screen-Recording-2023-05-29-at-1

### DIFF
--- a/backend/src/core/dto-mappers/letter.dto-mapper.ts
+++ b/backend/src/core/dto-mappers/letter.dto-mapper.ts
@@ -1,5 +1,5 @@
 import {
-  GetBulkLettersDto,
+  GetBulkLetterDto,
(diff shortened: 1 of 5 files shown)


gweiying (approved):
lgtm!

PR #136, opened 2023-05-30
chore: increase publicId length to 32

gweiying:
## Context

Increases length of publicId to 32 to make `publicId` more difficult to guess.


halfwhole (approved):
LGTM!

PR #137, opened 2023-05-30
feat: allow logins from whitelisted edu and health emails

halfwhole:
## Context

Letters should allow logins from whitelisted edu and health emails, not just gov.sg emails

Closes [Notion task](https://www.notion.so/opengov/Eng-Login-whitelist-7f8fe90e6efa41248ea6a3faf2c6f36f?pvs=4)

## Approach

Same approach as in this Highway PR: https://github.com/opengovsg/highway/pull/177

Updated the `IsGovSgEmail` guard to `IsGovSgOrWhitelistedEmail`, which matches the valid email glob expressions from for.sg and for.edu.sg.

Not the best long-term solution as every change to the email list will require a new release, but should suffice for now

## Deploy Notes

**New dependencies** (in shared):

- `minimatch`: for matching glob expressions


gweiying (approved):
lgtm!! thanks for implementing

PR #141, opened 2023-05-31
feat: template creation

disKeith:
## Context

This PR allows for users to create new templates using the TinyMCE editor

Closes [Notion Task](https://www.notion.so/opengov/Eng-Template-creation-52ae967d453f47c0ab54f5bb15d1b91c)

## Approach
Create a new page for users to create a new template with an editor, which upon saving - allows the user to generate and issue letters using the new template 

**Features**:

- `CreateTemplatePage.tsx`: page that renders the editor, allowing user to design and save a new template
- `CreateTemplateModal.tsx`: modal that creates a new template with specified `templateName` 
- `useCreateTemplateMutation`: custom hook to handle the mutation for creating a new template
- `{host}/admin/create`: route that renders the `CreateTemplatePage` component when `Add New` button is clicked
- `Button` in `TemplatesPage.tsx`: navigates user to template creation page when clicked
- `TemplateEditor`: TinyMCE editor component with plugins and tools for user to design a new template

**Bug Fixes**:

- Removed unused `useAdminAuth` and excess imports in `TemplatesPage.tsx` 

TODOs:
- `thumbnailS3Path` does not store the actual path for the thumbnail yet

## Videos

https://github.com/opengovsg/letters/assets/119388168/f1e8a145-9ca1-4020-bc8d-2314c1e9c5cd


frontend/src/features/create/components/CreateTemplateModal.tsx
```diff
@@ -0,0 +1,108 @@
...
+          <ModalCloseButton />
+          <ModalBody>
+            <FormControl isRequired>
+              <Input
+                {...register('templateName', {
+                  required: true,
+                  validate: validateName,
+                })}
+              />
+              {errors.templateName && (
```
gweiying, line 84: Is there a reason we use the manual `Text` component here instead of `FormErrorMessage`?

frontend/src/features/create/components/CreateTemplateModal.tsx
```diff
@@ -0,0 +1,108 @@
...
+            </FormControl>
+          </ModalBody>
+          <ModalFooter>
+            <Button variant="ghost" onClick={onClose} border={0}>
+              Cancel
+            </Button>
+            <Button
+              isLoading={isLoading}
+              color="#005DEA"
+              textColor="white"
```
gweiying, line 96: I think this styling might be unnecessary as we are using the default button colors!

frontend/src/features/create/components/CreateTemplateModal.tsx
disKeith: nope! thanks for catching this, updated in https://github.com/opengovsg/letters/pull/141/commits/e8ef970f7dd9455c95ef77914a857541b7fff9d3

frontend/src/features/create/components/CreateTemplateModal.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  const onSubmit = async (data: FormData): Promise<void> => {
+    await mutateAsync({
+      name: data.templateName.trim(),
+      fields: getFields(),
+      html: templateContent,
+      thumbnailS3Path: 'TODO',
+    })
+    onClose()
+    navigate(`/admin/${routes.admin.templates}`)
```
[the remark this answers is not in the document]
disKeith: oh cool didn't know that, thanks for this! updated in https://github.com/opengovsg/letters/pull/141/commits/e8ef970f7dd9455c95ef77914a857541b7fff9d3

frontend/src/features/create/components/CreateTemplateModal.tsx
disKeith: oh right, updated in https://github.com/opengovsg/letters/pull/141/commits/e8ef970f7dd9455c95ef77914a857541b7fff9d3

frontend/src/features/tinymce/components/Editor.tsx
```diff
@@ -42,3 +48,40 @@ export const Editor = ({
     </Box>
   )
 }
+
+export const TemplateEditor = ({
+  html,
+  onContentChange,
+  isDisabled = false,
+}: TemplateEditorProps): JSX.Element => {
+  const { tinymceApiKey, isLoadingTinymceApiKey } = useTinymceApiKey()
```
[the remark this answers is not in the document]
disKeith: ah, understood that the current `Editor` component was used for the letter view, so my initial thought was what you accurately mentioned and for readability - how about the following suggestions:

1. collapse into one `Editor` component and conditionally handle the return component
2. retain both components and update the `Editor` component to something like `LetterViewer`

after some thought into it, i'm leaning slightly towards option 1 as we'll avoid introducing duplicated code! wdyt?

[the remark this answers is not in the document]
gweiying: Synced offline, we plan to split the components into two since our public view, LetterViewer will unlikely use TinyMCE for rendering, so we will only use TinyMCE for internal rendering.

gweiying (approved):
lgtm! Pending design review from Pranu

~Note that we need to set up a separate API key for Tinymce on production, and set the approved domains as well.~
API key for production is setup and added to param store

PR #142, opened 2023-06-01
chore: update signup buttons to login

gweiying:
## Context

This PR updates the two buttons on the LandingPage to point to our OTP login page instead of our Sign Up beta form as the webapp is almost ready for testing. 

## Before & After Screenshots

**BEFORE**:
Screenshot 2023-06-01 at 9 32 04 AM

**AFTER**:
Screenshot 2023-06-01 at 9 31 03 AM


halfwhole (approved):
lgtm!

PR #145, opened 2023-06-01
Fix/render replace all

gweiying:
## Context
Clement discovered a bug where only the first appearance of a parameter gets replaced, instead of every parameter as we would expect. This is because `replace` only replaces the first appearance of the string matched, not every.

Changes:
- Changed from `replace` to `replaceAll`
- Add `"ES2021.String"` to the  Typescript compiler options as for some reason, `replaceAll` does not exist on type string otherwise
- Added tests to verify that all instances of a param replaced, not just the first one


halfwhole (approved):
LGTM

PR #148, opened 2023-06-07
fix: duplicate and case insensitive fields in template creation

disKeith:
## Context

When creating a new template, repeated fields (enclosed in double curly brackets) are added as new unique fields
## Approach

- Added shared
  - `TEMPLATE_KEYWORD_REGEX` to identify keywords in templates
  - `normalizeFields` to deduplicate and set fields to lowercase
  - `normalizeHtmlKeywords` to convert keywords in html to lowercase

- Frontend
  - Added logic to handle deduplicate and convert fields to lowercase inside the `getFields` function of `CreateTemplateModal`

- Backend
  - Added `processTemplate` function in `templates-parsing.service.ts` to store param cleaning and html parsing function
  - Updated `templates.service.ts` to call `processTemplate` before creating a new template

### DIFF
--- a/backend/src/templates/templates.controller.ts
+++ b/backend/src/templates/templates.controller.ts
@@ -27,16 +27,6 @@ export class TemplatesController {
 
   @Post()
   async create(@Body() templateDto: CreateTemplateDto) {
-    templateDto.fields = [
-      ...new Set<string>(
-        templateDto.fields.map((field: string) => field.toLowerCase()),
-      ),
-    ]
-    templateDto.html = templateDto.html.replace(
-      TEMPLATE_KEYWORD_REGEX,
-      (match: string, keyword: string) => `{{${keyword.toLowerCase()}}}`,
-    )
-
     return await this.templatesService
       .create(templateDto)
       .then((c) => mapTemplateToDto(c))
--- a/backend/src/templates/templates.controller.ts
+++ b/backend/src/templates/templates.controller.ts
@@ -10,6 +10,7 @@ import {
   UseGuards,
 } from '@nestjs/common'
 
+import { TEMPLATE_KEYWORD_REGEX } from '~shared/constants/regex'
 import {
   CreateTemplateDto,
   UpdateTemplateDto,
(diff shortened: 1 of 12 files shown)


backend/src/templates/templates.controller.ts
```diff
@@ -26,6 +27,16 @@ export class TemplatesController {
 
   @Post()
   async create(@Body() templateDto: CreateTemplateDto) {
+    templateDto.fields = [
```
gweiying, line 30: Quick comment, would be good to move this coercion/ validation logic into the service layer, because it's a specific business logic that the controller doesn't actually have to know. 

I think creating a shared util function to handle the parsing itself could also make sense, since it's a function that both the frontend and backend can use.

disKeith: oops! thanks, i have creating a separate service for it in https://github.com/opengovsg/letters/pull/148/commits/1c0fdafdb6a8568f1f6f80bc0a211e4a356909df

shared/src/util/templates.ts
```diff
@@ -0,0 +1,14 @@
+import { TEMPLATE_KEYWORD_REGEX } from '../constants/regex'
+
+export const normalizeFields = (fields: string[]): string[] => {
```
gweiying, line 3: nit, non-blocking! normalize doesn't seem like an intuitive name here, could we consider something else like `lowercase` or something more explicit?

disKeith: updated in https://github.com/opengovsg/letters/pull/148/commits/0bea9120bb4936b06f3fd85b7b303ecdc31f26c0

changed to `convertFieldsToLowerCase`, `deduplicateFields`, and `setHtmlKeywordsToLowerCase` - wdyt?

gweiying (approved):
lgtm!

frontend/src/features/create/components/CreateTemplateModal.tsx
```diff
@@ -47,13 +53,13 @@ export const CreateTemplateModal = ({
...
-    let match
-    while ((match = regex.exec(templateContent)) !== null) {
-      fields.push(match[1])
-    }
-    return fields
+    const fields: string[] = []
+    let match: RegExpExecArray | null
+
+    while ((match = TEMPLATE_KEYWORD_REGEX.exec(templateContent)) !== null)
+      if (!fields.includes(match[1])) fields.push(match[1])
```
gweiying, line 60: Just wondering, from my understanding this `if` condition is deduplicating the variables. But why do we have to deduplicate here and also in line 62?

gweiying (approved):
lgtm! just had one small question about the deduplication function, since it seems like we're doing it twice. I think it doesn't hurt so it's definitely non-blocking, was just curious more than anything

PR #149, opened 2023-06-08
Password protection (on bulk issueing)

AikoPath:
## Context

For Letters with sensitive information the users has asked to support password protection of letters, so that the letter can't be read by unauthorised parties.
Additionally to the password protection, this PR also encrypts each of the password protected letters (and the letter parameters) with the individual passwords as keys and does not store these keys so that only the issuing officer (the the recipient with the password) could decrypt the letter in the DB.

Closes [Notion: [Eng] Letter protection w/ password Issuing - Encryption](https://www.notion.so/opengov/Eng-Letter-protection-w-password-Issuing-Encryption-88f6df014b7646b7ba7aa43810f1377f?pvs=4)

Further details: [Notion: Letters Security](https://www.notion.so/opengov/Security-766d574464c94e0ca5c613a6b86f7cfe?pvs=4)

## Approach

- The officer provides the passwords in the csv
- The passwords are then extracted and send to the backend separately 
- On the backend the passwords are used to encrypt the letters and the letter parameters

**Improvements**:

- Refactors LettersValidationService. Adds unit test suite to ensure that it is working as expected.

- Removes unused column from `batches` table. (see details below)

- Removes toast that was hidden behind the bulk drawer.

## Before & After Screenshots

**BEFORE**:
Screenshot 2023-06-09 at 6 10 26 PM

**AFTER**:
Screenshot 2023-06-09 at 6 06 25 PM
Screenshot 2023-06-09 at 6 06 49 PM
Screenshot 2023-06-09 at 6 08 13 PM
Screenshot 2023-06-09 at 6 08 31 PM

## Deploy Notes

- `npm i` - to get crypto-js

**New migration scripts**:

- `1686218786357-remove-unused-rawcsv-from-batch`: Removes `rawCsv` from the `batches` table as it was unused and would have been needed to be encrypted unnecessarily for each individual row for password protected letters.

- `1686221927894-add-is-password-protected-flag-to-letter`: Adds a boolean field to the letters tabls to identify letters where the parameters and rendered letter are encrypted. This defaults to false for existing letters as non of them is encrypted.

**New dependencies**:

- `crypto-js` : used for encrypting (and later decrypting letters and )

## Risks

- This breaks if the officer does not want password protection but includes a `Password` field as an actual parameter in the csv file.

### DIFF
--- a/backend/src/core/dto-mappers/letter.dto-mapper.ts
+++ b/backend/src/core/dto-mappers/letter.dto-mapper.ts
@@ -39,6 +39,6 @@ export const mapLetterToGetBulkLetterDto = (
     ? getBulkLetterDtos
     : getBulkLetterDtos.map((dtos, index) => ({
         ...dtos,
-        Password: passwords[index],
+        Password: passwords[index], // Capitalized as it is what the user expects
       }))
 }
--- a/backend/src/core/dto-mappers/letter.dto-mapper.ts
+++ b/backend/src/core/dto-mappers/letter.dto-mapper.ts
@@ -27,10 +27,18 @@ export const mapLetterToDto = (letter: Letter): GetLetterDto => {
 export const mapLetterToGetBulkLetterDto = (
   letterParams: LetterParamMaps,
   letters: Letter[],
+  passwords?: string[],
 ): GetBulkLetterDto[] => {
-  return letters.map((letter, index) => ({
+  const getBulkLetterDtos = letters.map((letter, index) => ({
     ...letterParams[index],
     createdAt: letter.createdAt.toDateString(),
     publicId: letter.publicId,
   }))
+
+  return !passwords
+    ? getBulkLetterDtos
+    : getBulkLetterDtos.map((dtos, index) => ({
+        ...dtos,
+        Password: passwords[index],
+      }))
 }
--- a/backend/src/letters/letters-sanitization.service.ts
+++ b/backend/src/letters/letters-sanitization.service.ts
@@ -1,14 +1,15 @@
 import { Injectable } from '@nestjs/common'
 
-import { CreateLetterDto } from '~shared/dtos/letters.dto'
 import { sanitizeHtml } from '~shared/util/html-sanitizer'
+
+import { RenderedLetter } from './letters-rendering.service'
 
 @Injectable()
 export class LettersSanitizationService {
-  sanitizeLetter(createLetterDto: CreateLetterDto): CreateLetterDto {
+  sanitizeLetter(letter: RenderedLetter): RenderedLetter {
     return {
-      ...createLetterDto,
-      issuedLetter: sanitizeHtml(createLetterDto.issuedLetter),
+      ...letter,
+      issuedLetter: sanitizeHtml(letter.issuedLetter),
     }
   }
 }
--- a/backend/src/letters/letters-validation.service.spec.ts
+++ b/backend/src/letters/letters-validation.service.spec.ts
@@ -22,8 +22,14 @@ describe('LettersValidationService', () => {
 
     it('should validate passwords when they are enabled but none are set', () => {
       const passwords: string[] = ['']
+      const fields = ['field1']
+      const letterParamMaps: LetterParamMaps = [
+        {
+          field1: 'param1',
+        },
+      ]
 
-      const result = service.validateBulk([], [], passwords)
+      const result = service.validateBulk(fields, letterParamMaps, passwords)
 
       expect(result.success).toBe(false)
       expect(result.message).toEqual('Malformed bulk create object')
@@ -34,8 +40,20 @@ describe('LettersValidationService', () => {
 
     it('should fail when one password is missing ', () => {
       const passwords: string[] = ['hunter2', '', 'hunter2']
+      const fields = ['field1']
+      const letterParamMaps: LetterParamMaps = [
+        {
+          field1: 'param1',
+        },
+        {
+          field1: 'param1',
+        },
+        {
+          field1: 'param1',
+        },
+      ]
 
-      const result = service.validateBulk([], [], passwords)
+      const result = service.validateBulk(fields, letterParamMaps, passwords)
 
       expect(result.success).toBe(false)
       expect(result.message).toEqual('Malformed bulk create object')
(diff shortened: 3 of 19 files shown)


backend/src/core/dto-mappers/letter.dto-mapper.ts
```diff
@@ -27,10 +27,18 @@ export const mapLetterToDto = (letter: Letter): GetLetterDto => {
...
     ...letterParams[index],
     createdAt: letter.createdAt.toDateString(),
     publicId: letter.publicId,
   }))
+
+  return !passwords
+    ? getBulkLetterDtos
+    : getBulkLetterDtos.map((dtos, index) => ({
+        ...dtos,
+        Password: passwords[index],
```
gweiying, line 42: is there a small typo here? think it should be `password` instead of `Password`?

backend/src/letters/letters.service.ts
```diff
@@ -44,50 +46,60 @@ export class LettersService {
...
     const renderedLetters = this.lettersRenderingService.bulkRender(
       template.html,
       letterParamMaps,
     )
+
+    const sanitizedLetters = renderedLetters.map((renderedLetter) =>
+      this.lettersSanitizationService.sanitizeLetter(renderedLetter),
+    )
+
+    const secureLetters = !passwords
```
gweiying, line 79: tiniest suggestion, since we're using the past tense for all our variables above e.g. `rendered`, `sanitized`, should this be `secured`?

backend/src/core/dto-mappers/letter.dto-mapper.ts
gweiying: Oops, realized that it's correctly `Password` since this is being returned to the frontend. Perhaps small suggestion, can we add a small comment after?

AikoPath: Makes sense, added a comment.

backend/src/letters/letters.service.ts
AikoPath: 👍
`secured` it is

backend/src/letters/letters-validation.service.ts
```diff
@@ -13,47 +13,30 @@ export class LettersValidationService {
...
-          letterParamMap[field] === ''
-        ) {
-          errorArray.push({
-            id: i,
-            param: field,
-            message: BulkLetterValidationResultErrorMessage.MISSING_PARAM,
-          })
-        }
-      }
+    const errors: BulkLetterValidationResultError[] = []
```
gweiying, line 25: Thanks for this refactor! looks much cleaner

backend/src/letters/letters.service.ts
```diff
@@ -44,50 +46,60 @@ export class LettersService {
...
-        rawCsv: JSON.stringify(letterParamMaps),
       } as CreateBatchDto
 
       const batch = await this.batchesService.createWithTransaction(
         createBatchDto,
         entityManager,
       )
-      const lettersDto = renderedLetters.map(
+      const lettersDto = secureLetters.map(
         (renderedLetter: Partial<Letter>) => ({
```
gweiying, line 96: nit, do we want to change this to `securedLetter` or `letter`, instead of `secureLetters.map(
        (renderedLetter: Partial<Letter>)`,

AikoPath: done ✅

backend/src/letters/letters-validation.service.ts
```diff
@@ -62,7 +45,62 @@ export class LettersValidationService {
     return {
       success: false,
       message: 'Malformed bulk create object',
-      errors: errorArray,
+      errors: errors,
```
gweiying, line 48: nit, javascript syntax, it's sufficient to return this
```
return {
      success: false,
      message: 'Malformed bulk create object',
      errors,
}
```

AikoPath: Good catch, Thanks!
I must have overlooked that when automatically refactoring.
Changed.

backend/src/letters/letters-validation.service.spec.ts
```diff
@@ -0,0 +1,233 @@
...
+
+    it('should succeed when all passwords are provided ', () => {
+      const passwords: string[] = ['hunter2', 'hunter2', 'hunter2']
+
+      const result = service.validateBulk([], [], passwords)
+
+      expect(result.success).toBe(true)
+      expect(result.message).toEqual('Validation Success')
+      expect(result.errors).toBeUndefined()
+    })
```
gweiying, line 55: Just wondering, currently we don't have any checks to make sure that the password array is same length as the `letterParamMaps`, would that be a problem?

frontend/src/features/issue/BulkIssueDrawer/UploadCsvForm.tsx
```diff
@@ -113,7 +156,8 @@ export const UploadCsvForm = ({
             isDisabled={
               !(parsedArr.length > 0) ||
               !!parseCsvError ||
-              uploadCsvErrors.length > 0
+              uploadCsvErrors.length > 0 ||
+              (isPasswordProtected !== passwords.length > 0 && !!file)
```
gweiying, line 160: just wondering if this `!!file` check can be moved out, i.e.
```
isDisabled={
  !!file ||
  !(parsedArr.length > 0) ||
  !!parseCsvError ||
  uploadCsvErrors.length > 0 ||
  isPasswordProtected !== passwords.length > 0
}
```

frontend/src/features/issue/BulkIssueDrawer/UploadCsvForm.tsx
```diff
@@ -40,34 +42,58 @@ export const UploadCsvForm = ({
...
+      : { templateId, letterParamMaps: parsedArr }
+    await mutateAsync(reqBody)
+  }
+
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setIsPasswordProtected(event.target.checked)
+  }
+
+  const getErrorMessage = (): string => {
+    let errorMessage = ''
```
gweiying, line 76: tiny suggestion, also non-blocking,

should we simplify the `file` check to return early, for example by adding this line,
`if (!file) return ''`

gweiying (approved):
looks good to me! left some inline comments but the bulk are mostly nits and non-blocking, thanks for the PR!

backend/src/letters/letters-sanitization.service.ts
```diff
@@ -1,14 +1,15 @@
...
+import { RenderedLetter } from './letters-rendering.service'
+
 @Injectable()
 export class LettersSanitizationService {
-  sanitizeLetter(createLetterDto: CreateLetterDto): CreateLetterDto {
+  sanitizeLetter(letter: RenderedLetter): RenderedLetter {
     return {
-      ...createLetterDto,
-      issuedLetter: sanitizeHtml(createLetterDto.issuedLetter),
+      ...letter,
```
gweiying, line 11: Small nit, but might be outside the scope of this PR, we might want to move this `...letter` outside of this function so that it's more clear what `sanitizeLetter` is returning

frontend/src/features/issue/BulkIssueDrawer/UploadCsvForm.tsx
gweiying: something like this might also be easier to read

```
const getErrorMessage = (): string => {
    if (!file) return ''
    if (parseCsvError) {
      return parseCsvError
    } 
    if (!isPasswordProtected && passwords.length > 0) {
      return 'Password field found in the CSV file despite Password protection disabled, please upload an updated .csv'
    }
    if (isPasswordProtected && passwords.length === 0) {
      return 'No Password field found in the CSV file despite Password protection enabled, please upload an updated .csv'
    }
  }
```

backend/src/letters/letters-validation.service.spec.ts
AikoPath: Just to be 100% certain I have added a check for this. Given our current Frontend this can never happen though, but I guess it's good to have this to catch regression.

frontend/src/features/issue/BulkIssueDrawer/UploadCsvForm.tsx
AikoPath: Yes,  with a `!file` instead of `!!file` this is nice, thanks.

gweiying (approved):
lgtm!

PR #154, opened 2023-06-15
View password protected letters

AikoPath:
## Context

Since the officers are now able to password protect letters, this PR enables the view for MoP to enter a password and then view the decrypted letter.

Closes [Notion: [Eng] Letter protection w/ password Viewing - Decryption](https://www.notion.so/opengov/Eng-Letter-protection-w-password-Viewing-Decryption-ff71bb02d2d34e8c846eae24bd31ffd0?pvs=4)

## Approach

- Re-use existing GET request to query for the letter
  - In case of non protected letters the existing Letter view (and retrieval) does not change
- For password protected letters the response from the BE is an error with HTTP status 401 when no password is provided
  - Upon receiving the 401 error the password field is presented to the MoP
  - When a password is entered by the MoP, this password is sent to the BE as a `password` header.
  - If that password is the valid key to decrypt the letter, that letter is returned and shown to the MoP (as if it were)
  - If that password is not a valid key to decrypt the letter, an error with HTTP status 401 and and error message "Invalid Password" is returned and shown to the MoP.

**Features**:

- View password protected Letters
- Mobile friendly
- Keep existing non-password-protected letter unaffected.

**Improvements**:

- Moves `crypto-js` dependencies from top level `package.json` to backend `package.json`
- Adds sharable link to the Postman collection to make it easier to keep the collection in sync (this should eventually make sharing the collection via this repo unnecessary).
- Rename `encryptLetters` to `encryptLetter` since it is encrypting one letter.
- Removed some unused code (mostly imports)

**Bug Fixes**:

- Hides "Download PDF" button when letter is still loading

## Before & After Screenshots

**BEFORE**:
Realistically the password protection feature did not exist before so there is no **before** screenshot.
Technically however, the password protection was enabled with the last PR and this was the intermediary view for password protected letter prior to this PR.
Screenshot 2023-06-15 at 2 12 48 PM

**AFTER**:
Screen Recording 2023-06-15 at 4 55 04 PM

**Password View** 
Screenshot 2023-06-15 at 5 15 02 PM
**Password View with Invalid Password**
Screenshot 2023-06-15 at 5 15 11 PM
**Mobile View ports**
Screen Recording 2023-06-15 at 5 17 01 PM

**Questions**:
- Are we using or are we intending to use the GET '/public/:publicId' endpoint on the letters controller?


backend/src/letters/letters.service.ts
```diff
@@ -151,4 +158,16 @@ export class LettersService {
   remove(id: number) {
     return `This action removes a #${id} letter`
   }
+
+  private decryptLetter(password: string | undefined, letter: Letter) {
```
gweiying, line 162: nit, 

wondering if this functionality can be moved into the LettersEncryptionService as it feels slightly unintuitive that there's another wrapper function to be called from within LettersService to decrypt the letter? What if we tried something like this within LettersEncryptionService:

```
decryptLetter(letter: Letter, password: string): Letter {
    if (!password) { 
      throw new UnauthorizedException('No Password provided')
    }

    try {
      return this.decrypt(letter, password)
    } catch (error) {
      throw new UnauthorizedException('Invalid Password')
    }
  }

  private decrypt(letter: Letter, password: string): Letter {
    const fieldValues = AES.decrypt(letter.fieldValues, password).toString(
      enc.Utf8,
    )
    const issuedLetter = AES.decrypt(letter.issuedLetter, password).toString(
      enc.Utf8,
    )

    // For short input and wrong decryption key AES removes all output (while removing padding).
    // More details: https://go.gov.sg/aes-padding
    if (fieldValues === '' && issuedLetter === '') {
      throw new Error('Invalid Key')
    }

    return {
      ...letter,
      fieldValues,
      issuedLetter,
    }
  }
```
Also swapped the order of the variables `'password', 'letter'` around here, as `LettersService.decryptLetter(password: string | undefined, letter: Letter)` has a different order from `LettersEncryptionService.decryptLetter(letter, password)`, which I think was unintentional?

frontend/src/features/auth/LoginPage.tsx
```diff
@@ -51,17 +52,8 @@ const BaseGridLayout = (props: GridProps) => (
...
-    gridColumn={{ base: '1 / 5', md: '2 / 12', lg: '8 / 12' }}
-    py="4rem"
-    display="flex"
-    alignItems={{ base: 'initial', lg: 'center' }}
-    justifyContent="center"
-  >
-    {children}
-  </GridItem>
-)
+const LoginGridArea: FC<PropsWithChildren> = ({ children }) =>
```
gweiying, line 55: Thanks for cleaning this up!

frontend/src/features/public/LetterPublicPage.tsx
```diff
@@ -1,47 +1,80 @@
...
 
   const handleDownload = () => {
     if (letter && letter.issuedLetter) {
       void convertHtmlToPdf(letter.issuedLetter, `${letterPublicId}.pdf`)
     }
   }
 
-  if (!isLetterLoading && !letter) {
+  const handleSubmit = async (event: FormEvent<HTMLFormElement>) => {
+    event.preventDefault()
```
gweiying, line 34: Just curious, why is this line above necessary?

AikoPath: This prevents the page from reloading upon form submission.

backend/src/letters/letters.service.ts
AikoPath: Good catch on the argument order! 👏

I also agree on pulling this into the encryption service.
Addressing this in the next commit.

backend/src/letters/letters.controller.ts
```diff
@@ -75,7 +75,7 @@ export class LettersController {
     return this.lettersService.findOne(+id)
   }
 
-  @Get('/public/:publicId')
+  @Get('/public/:publicId') // Do we still use/need this?
```
gweiying, line 78: Good point!

frontend/src/features/public/LetterPublicPage.tsx
gweiying: Got it, thanks. Small suggestion, completely non-blocking. What do you think of moving this `event.preventDefault()` line into the `PasswordProtectedView` component, under `handleSubmit`? Might be more intuitive there since the event is being generated there from the `<form>`

backend/src/letters/letters-encryption.service.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+        password,
+      )
+
+      expect(encryptedLetter.issuedLetter).not.toEqual(letter)
+      expect(decryptedLetter.issuedLetter).toEqual(letter)
+      expect(encryptedLetter.fieldValues).not.toEqual(letterParamMap)
+      expect(decryptedLetter.fieldValues).toEqual(letterParamMap)
+    })
+
+    it('should NOT decrypt a letter when the provided password is wrong (for a short letter)', () => {
```
gweiying, line 34: Should we add a test for an empty password string?

backend/src/letters/letters-encryption.service.ts
```diff
@@ -1,14 +1,36 @@
...
 export class LettersEncryptionService {
-  encryptLetters(letter: RenderedLetter, password: string): RenderedLetter {
+  encryptLetter(letter: RenderedLetter, password: string): RenderedLetter {
     return {
       fieldValues: AES.encrypt(letter.fieldValues, password).toString(),
       issuedLetter: AES.encrypt(letter.issuedLetter, password).toString(),
     }
   }
+
+  decryptLetter(letter: Letter, password: string): Letter {
```
gweiying, line 16: Left a comment above about abstracting the error throwing here

frontend/src/features/public/LetterPublicPage.tsx
AikoPath: Good idea!
Done.

backend/src/letters/letters-encryption.service.spec.ts
AikoPath: As this is not a valid use case I don't think we should have a test for it. It could give and invalid use case unjustified legitimacy.
AES itself (and the crypto-js implementatino) handles this case of an empty string correctly.

backend/src/letters/letters.controller.ts
AikoPath: Removed this.

backend/src/letters/letters-encryption.service.ts
```diff
@@ -1,14 +1,44 @@
...
+
+  private decrypt(letter: Letter, password: string): Letter {
+    const fieldValues = AES.decrypt(letter.fieldValues, password).toString(
+      enc.Utf8,
+    )
+    const issuedLetter = AES.decrypt(letter.issuedLetter, password).toString(
+      enc.Utf8,
+    )
+
+    // For short input and wrong decryption key AES removes all output (while removing padding).
```
gweiying, line 32: Just a small suggestion here, also non-blocking. I think it would also be good to mention in the inline comment the other possibility that for longer input and wrong decryption keys, the wrong decryption would throw a `Malformed UTF-8 data` error.

gweiying: Thinking harder about the decryption process, I think there's a scenario in which decrypting with a wrong key might not result in an error, but just outputs a wrong decryption string. Will catch you offline to discuss this

gweiying: Chatted offline, will not handle for now as the chances are minuscule and not worth handling for now.

gweiying (approved):
lgtm! left a small suggestion about documentation, but non-blocking! thanks for the PR

PR #155, opened 2023-06-15
fix: fix zoom on hover and border radius of template card

AikoPath:
## Context

- The zoom animation of the template card was broken as that card was cut off on any side of the parent container when zoomed in
- Th shadow on any side touching the parent container was cut off.
- The top border radius of the card was not correct.

This PR fixes these issues.

## Before & After Screenshots

**BEFORE**:
Screen Recording 2023-06-15 at 7 12 11 PM

**AFTER**:
Screen Recording 2023-06-15 at 7 13 07 PM


gweiying (approved):
lgtm!

PR #162, opened 2023-06-23
chore: add OTP_EXPIRY to letters ecs-task-definition

gweiying:
We want to increase the default OTP expiration time. As we need to set the value via the Param Store, we need to add it to the `ecs-task-definition.json` file so that the value will be read from Param Store on deployment.


AikoPath (approved):
lgtm

PR #163, opened 2023-06-23
fix: remove empty lines from CSV processing

PikkaPikkachu:
## Context
We are trying to make the CSV uploads flow smoother, by eliminating blank lines from the CSV processing automatically instead of throwing error to the user that there is a blank line missing, which can sometimes be hard for the user to debug since, they are unable to recognise or "see" the missing lines on excel. 

Hence the code change involved here, is to remove the blank lines from CSV and also inform the user the number of blank lines that were ignored while their CSV file was processed.

Closes this [notion task](https://www.notion.so/opengov/Eng-Error-handling-for-CSV-uploads-47eb937d69274a03b4ae551446d16f4c)

## Approach

Make changes to the `papaparse` config to ignore the empty lines in the CSV.

We have implemented 1 in this PR.

## Before & After Screenshots

**BEFORE**:
## When a missing row was present 
--> error were being thrown
Screenshot 2023-06-23 at 6 28 29 PM

**AFTER**:
Tested for three different test cases:

## When the blank rows are ignored in the CSV

Screenshot 2023-06-26 at 3 48 00 PM

## When blank rows are zero, no blank rows ignored
Screenshot 2023-06-23 at 6 11 06 PM

## When there's some usual error [regression testing]
Screenshot 2023-06-23 at 6 06 41 PM

### DIFF
--- a/frontend/src/utils/csvUtils.ts
+++ b/frontend/src/utils/csvUtils.ts
@@ -48,6 +48,7 @@ export const csvToJsonArr = async (file: File): Promise<unknown[]> => {
       complete: (results) => {
         resolve(results.data)
       },
+      skipEmptyLines: true,
       error: (error) => {
         reject(error)
       },


AikoPath (changes requested):
Thanks for this PR.
Based on the team discussion this morning we don't actually need to show how many empty lines we have removed.

AikoPath (approved):
LGTM

PR #164, opened 2023-06-27
feat: Add QR Codes to Letters 

PikkaPikkachu:
## Context

We are adding QR codes to letters, so that anyone can scan them and easily access the the link to the letter. 

Closes [this notion task](https://www.notion.so/opengov/Eng-Dynamic-QR-code-generation-ba96a9971dea42b48005f4f1862b5e46)

## Approach

We are using `react-qr-code` library for generating QR codes for links to different letters on client side. 
More details on the approach can be found [here](https://www.notion.so/opengov/20230624-QR-Code-Generation-for-Letters-b36e2699faff4a40b72279fb467b4c44)

**Features**:

- QR code will be shown when the letter is viewed and if there's a letter ID that exists. 
- No QR code is shown in preview [that's because we don't have a letter ID, although we can show a sample QR code for just preview purposes if required.]

## Before & After Screenshots

**BEFORE**:
Letter preview without the QR code:

Screenshot 2023-06-27 at 2 02 35 PM

**AFTER**:

Flow of where and how the QR codes are seen after the new changes:

https://github.com/opengovsg/letters/assets/25703976/952959bb-94db-4e0b-b9ae-82d0660f436e

**New dev dependencies**:

`"react-qr-code": "^2.0.11"`

### DIFF
--- a/frontend/src/features/editor/components/LetterQRCode.tsx
+++ b/frontend/src/features/editor/components/LetterQRCode.tsx
@@ -1,4 +1,4 @@
-import { Text } from '@chakra-ui/react'
+import { Box } from '@chakra-ui/react'
 import QRCode from 'react-qr-code'
 
 interface LetterQRCodeProps {
@@ -9,28 +9,16 @@ export const LetterQRCode = ({
   qrCodeLinkValue,
 }: LetterQRCodeProps): JSX.Element => {
   return (
-    <div style={{ margin: 'auto', padding: '25px' }}>
-      <hr />
-      <Text>
-        This letter was retrieved from <b>{qrCodeLinkValue}</b>
-      </Text>
-      <div
+    <Box w="10%" h="10%">
+      <QRCode
+        size={12}
         style={{
-          height: '25%',
-          width: '25%',
-          margin: 'auto',
+          height: 'auto',
+          maxWidth: '100%',
+          width: '100%',
         }}
-      >
-        <QRCode
-          size={12}
-          style={{
-            height: 'auto',
-            maxWidth: '100%',
-            width: '100%',
-          }}
-          value={qrCodeLinkValue}
-        />
-      </div>
-    </div>
+        value={qrCodeLinkValue}
+      />
+    </Box>
   )
 }
--- a/frontend/src/features/editor/components/LetterQRCode.tsx
+++ b/frontend/src/features/editor/components/LetterQRCode.tsx
@@ -1,0 +1,36 @@
+import { Text } from '@chakra-ui/react'
+import QRCode from 'react-qr-code'
+
+interface LetterQRCodeProps {
+  qrCodeLinkValue: string
+}
+
+export const LetterQRCode = ({
+  qrCodeLinkValue,
+}: LetterQRCodeProps): JSX.Element => {
+  return (
+    <div style={{ margin: 'auto', padding: '25px' }}>
+      <hr />
+      <Text>
+        This letter was retrieved from <b>{qrCodeLinkValue}</b>
+      </Text>
+      <div
+        style={{
+          height: '25%',
+          width: '25%',
+          margin: 'auto',
+        }}
+      >
+        <QRCode
+          size={12}
+          style={{
+            height: 'auto',
+            maxWidth: '100%',
+            width: '100%',
+          }}
+          value={qrCodeLinkValue}
+        />
+      </div>
+    </div>
+  )
+}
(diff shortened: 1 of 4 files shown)


gweiying:
Implementation looks ok, but could you check with Pranu on the design of this, since we also have another section below that has the letter public id?

AikoPath:
I agree with @gweiying.

I think we need a little bit cleaner design for this.
Maybe we can find something that looks a little nicer than the digiMC one.
Screenshot 2023-06-27 at 6 09 21 PM

PikkaPikkachu:
Discussed with Pranu and updated the design for QR codes.

frontend/src/features/editor/components/LetterViewer.tsx
```diff
@@ -28,7 +30,9 @@ export const LetterViewer = ({
         borderColor="base.divider.medium"
         bg="white"
       >
-        <div dangerouslySetInnerHTML={{ __html: cleanHtml }}></div>
+        <>
```
gweiying, line 33: Nit, non-blocking, this wrapper seems to be unused now?

frontend/src/features/editor/components/LetterViewer.tsx
```diff
@@ -40,15 +44,22 @@ export const LetterViewer = ({
...
-              This letter is generated using {document.location.host}. Forging
-              or morphing is a criminal offence.
-            </Text>
-          </VStack>
+          <HStack spacing="0.25rem" alignItems="flex-start">
+            <LetterQRCode
+              qrCodeLinkValue={`${document.location.host}/letters/${letterPublicId}`}
+            />
+            <div style={{ margin: 'auto 10px' }}>
+              <VStack spacing="0.25rem" alignItems="flex-start">
```
gweiying, line 52: nit, consider collapsing the styling in the `div` into the VStack. Since your `div` is just wrapping a single child `VStack`, it probably means you don't need the `div` and can add the styling directly into the `VStack` to avoid unnecessary nesting. 

Might also be personal preference, but I would shy away from mixing Chakra components with pure html. For example, the Chakra `Box` component returns a `div`, so instead of using a `div` component, you could use a `Box` instead. This would probably have some benefits apart from code cleanliness, like ease of style sharing for example.

frontend/src/features/editor/components/LetterQRCode.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+interface LetterQRCodeProps {
+  qrCodeLinkValue: string
+}
+
+export const LetterQRCode = ({
+  qrCodeLinkValue,
+}: LetterQRCodeProps): JSX.Element => {
+  return (
+    <div
```
gweiying, line 12: Nit and totally non-blocking because there's no functionality impact and we're not actually using any other chakra-ui components in this page anyway, but to give an example of replacing the div with Chakra components, this is what it might look like:
```
<Box w="10%" h="10%">
  <QRCode
      size={12}
      style={{
        height: 'auto',
          maxWidth: '100%',
        width: '100%',
      }}
      value={qrCodeLinkValue}
    />
</Box>
```

frontend/src/features/editor/components/LetterQRCode.tsx
```diff
@@ -0,0 +1,29 @@
+import { Text } from '@chakra-ui/react'
```
gweiying, line 1: Unused import

gweiying (commented):
Functionality and styling looks much better, thanks for checking in with Pranu on this. Left some comments in line on the frontend component styling, nothing major.

Just wondering if we might also want to add a small copy explaining that scanning the QR code sends them back to this letter, or would people wonder what the QR code is for? I can also raise the question to the wider team if you like, not sure if the confusion is mostly just mine.

PikkaPikkachu:
Thanks Alexis, yeah we could check with the team for whether we wanna add text explaining the significance of QR — I think we partially discussed this with Pranu as well on Wednesday, but would be good to confirm! :)

frontend/src/features/editor/components/LetterViewer.tsx
PikkaPikkachu: Oops, seems like I forgot to remove it, after the design fixes, thank you!

frontend/src/features/editor/components/LetterViewer.tsx
PikkaPikkachu: Makes sense!

AikoPath (approved):
LGTM.

And yes, the design looks waay better now, thanks!

PR #166, opened 2023-06-27
Feat/send via sms

AikoPath:
## Context

We want to allow officers to contact citizens with their issued letters. This PR implements the SMS notification flow, so that officers can trigger SMSes to citizens with their issued letter links.

## Approach

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:
[insert screenshot here]

**AFTER**:
[insert screenshot here]

## Tests

## Deploy Notes

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:

## Risks


backend/src/types/notifications.ts
```diff
@@ -0,0 +1,5 @@
+export enum communicationChannels {
```
thanhdatle, line 1: small nit: enum should be singular

backend/src/database/entities/notifications.entity.ts
```diff
@@ -0,0 +1,39 @@
...
+  // changing this requires a db migration
+  // do we need to settle on this now or can we decide this later and not have a column/psql enum for now?
+  // is it even worth having an PSQL enum for this in general?
+  @Column({
+    type: 'enum',
+    enum: communicationChannels,
+  })
+  channel: communicationChannels
+
+  @Column('int')
```
gweiying, line 25: Missing foreign key relation

backend/src/database/entities/notifications.entity.ts
```diff
@@ -0,0 +1,39 @@
...
+  letter_id: number
+
+  @Column('varchar', { length: 255, nullable: false, unique: true })
+  uuid: string
```
gweiying, line 29: small question - `uuid` makes it sound like an id that we're generating, but it's actually a unique id that we receive from the provider - possibly something like `providerId` might encapsulate this better?

backend/src/database/entities/notifications.entity.ts
```diff
@@ -0,0 +1,39 @@
...
+
+@Entity({ name: 'notifications' })
+export class Notifications {
+  @PrimaryGeneratedColumn()
+  id: number
+
+  // Todo: are we sure these are the channels we will support?
+  // changing this requires a db migration
+  // do we need to settle on this now or can we decide this later and not have a column/psql enum for now?
+  // is it even worth having an PSQL enum for this in general?
```
gweiying, line 18: good point, let's discuss this over the call!

backend/src/config/config.schema.ts
```diff
@@ -244,4 +245,18 @@ export const schema: Schema<ConfigSchema> = {
...
+  twilioApiKey: {
+    accountSid: {
+      doc: 'Account SID used for Twilio',
+      env: 'TWILIO_SID',
+      format: String,
+      default: '',
+    },
+    authToken: {
+      doc: 'Auth Token used for Twilio',
+      env: 'TWILIO_AUTH_TOKEN',
```
gweiying, line 257: need to read this in ecs-task-definition.json

backend/src/database/entities/notifications.entity.ts
gweiying: Agreed that it makes more sense to use Typescript enums but not enforce at DB level since any changes will require further migrations as mentioned above!

backend/src/letters/letters-notifications.service.ts
```diff
@@ -0,0 +1,96 @@
...
+          const messageSid = await this.twilioService.sendMessage(
+            message,
+            recipient,
+          )
+          messageSids.push(messageSid)
+        } catch (error) {
+          throw { success: false }
+        }
+      }),
+    )
```
gweiying, line 74: Just wondering if we might end up with arrays of different length, where `messageSids` would be shorter than `messages` if the message fails to send

gweiying: Discussed asynchonously - `allSettled` makes sense. We should probably (assuming Twilio's latency is reasonable) return message with `x smses successfully sent` or `y messages failed to send` and allow users to retry

backend/src/letters/letters-notifications.service.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  private async createBulkNotification(
+    channel: communicationChannels,
+    letters: Letter[],
+    uuid: string[],
+    messages: string[],
+    recipients: string[],
+  ): Promise<Notifications[]> {
+    const newNotifications: Notifications[] = letters.map((letter, index) => {
```
gweiying, line 85: Notification table should record all messages even if Twilio fails to schedule it. So we will probably need a separate `status` column, that records whether it was scheduled successfully on Twilio or not.

gweiying: We decided to use the status column instead of using the lack of a `uuid` because it's more explicit and easier to track?

backend/src/types/notifications.ts
gweiying: nit, `CommunicationChannel`

backend/src/letters/letters-notifications.service.ts
```diff
@@ -0,0 +1,96 @@
...
+    const newNotifications: Notification[] = letters.map((letter, index) => {
+      const notification = new Notification()
+      notification.channel = channel
+      notification.letterId = letter.id
+      notification.providerId = uuid[index]
+      notification.message = messages[index]
+      notification.recipient = recipients[index]
+      return notification
+    })
+    return await this.notificationRepository.save(newNotifications)
```
gweiying, line 94: Might be good to look into bulk insertion instead

AikoPath: Oh, does TypeORM not do a bulk insert when save is called with an array of entities?

gweiying: Oops yes it does.

backend/src/database/entities/notification.entity.ts
```diff
@@ -0,0 +1,37 @@
...
+  letterId: number
+
+  @Column('varchar', { length: 255, unique: true, nullable: true })
+  providerId: string
+
+  @Column('text')
+  message: string
+
+  @Column('text')
+  recipient: string
```
PikkaPikkachu, line 33: n00b question — what does recipient represent here? Is the recipient's name or some ID?

backend/src/letters/letters-notifications.service.ts
gweiying: Update, we will not be using the status column for now as it's additional complexity on the table that we don't need for now.

backend/src/database/entities/notification.entity.ts
AikoPath: The phone numnber of the recipient.
You can check the twilio service test to sww how it's used.

backend/src/letters/letters-validation.service.spec.ts
```diff
@@ -13,7 +13,7 @@ describe('LettersValidationService', () => {
     it('should not validate passwords if none are set', () => {
       const passwords = undefined
 
-      const result = service.validateBulk([], [], passwords)
+      const result = service.validateBulk([], [], passwords, undefined)
```
PikkaPikkachu, line 16: minor nit: I think we should explicitly comment or define here what the `undefined` means? And same for all function calls below — will be easier to debug in future?

backend/src/letters/letters.service.spec.ts
```diff
@@ -1,44 +0,0 @@
-import { Test, TestingModule } from '@nestjs/testing'
```
PikkaPikkachu, line 1: hmm curious as to why we deleted this file?

backend/src/notifications/notifications.service.ts
```diff
@@ -0,0 +1,53 @@
...
+  constructor(
+    @InjectRepository(Notification)
+    private readonly notificationRepository: Repository<Notification>,
+    private readonly smsNotificationService: SmsNotificationsService,
+    @InjectRepository(User)
+    private readonly usersRepository: Repository<User>,
+    private readonly config: ConfigService,
+  ) {}
+
+  async sendNotifications(
```
PikkaPikkachu, line 20: minor nit: should this be called `sendSMSNotification` so that it's super easy to understand what kind of notification we are talking about?

backend/src/notifications/notifications.service.ts
```diff
@@ -0,0 +1,53 @@
...
+    userId: number,
+    letters: Letter[],
+    recipients: string[],
```
PikkaPikkachu, line 23: minor nit: Hmm the recipients variable is a bit confusing, can't we say something like `recipientPhoneNumbers` so that we know exactly what kind of data is present in the array, right now that's not super clear?

backend/src/letters/letters.service.spec.ts
AikoPath: It is not testing anything (-> not bringing any value) but required effort to keep updated as the dependencies on the service were changed.

backend/src/types/notification.ts
```diff
@@ -0,0 +1,5 @@
+export enum NotificationChannel {
+  SINGPASS = 'SINGPASS',
+  TWILIO = 'TWILIO',
+  POSTMAN = 'POSTMAN',
```
AikoPath, line 4: I think we should limit this to only the channel we actually have right now i.e. `TWILIO`

backend/src/database/entities/notification.entity.ts
```diff
@@ -0,0 +1,37 @@
...
+  @Column('text')
+  channel: string
+
+  @ManyToOne(() => Letter)
+  letter: Letter
+  @Column('int')
+  letterId: number
+
+  @Column('varchar', { length: 255, unique: true, nullable: true })
+  providerId: string
```
thanhdatle, line 27: I am having second thought on the way we record error to the providerId column. Since we make this column unique and have a fixed length, we have to append some random string to the error message. This will make it hard/weird logic to look for failed notifications.
What is the implication of removing unique/fixed string length constraints?

shared/src/constants/regex.ts
```diff
@@ -1 +1,4 @@
 export const ACCEPTED_TEMPLATE_FIELDS_REGEX = /{{( *[a-zA-Z0-9_]+ *)}}/g
+
+export const ACCEPTED_SINGAPORE_PHONE_NUMBERS_REGEX =
```
thanhdatle, line 3: since we are writing unit tests, should we include some tests to test the regex since its readability isn't great

AikoPath: There should be somes tests for this in the `validation.service.spec`

backend/src/database/entities/notification.entity.ts
AikoPath: I think the idea was to keep this column empty for error cases. (multiple Null values should be allowed even when a uniqueness constraint exists), and just log(?) the error.
That being said, removing the fixed length should be no problem at all. The uniqueness constraint also does not offer a lot of value tbh.

AikoPath:
I can't actully approve this PR (as I have opened this :) )
But it mostly looks good to me.
Thanks a lot @gweiying and @disKeith for getting this done! 🌮🌮🌮

shared/src/constants/regex.ts
thanhdatle: got ya

thanhdatle:
Not sure if this is too late to ask this question but should we include a status field ('CREATED', 'SENT', 'SENT_FAILED', 'READ') into letter.entity.ts?
I guess we can derive the value from the Notification's providerID and letter's firstReadAt? What are your thoughts on this? @AikoPath @gweiying

AikoPath:
> Not sure if this is too late to ask this question but should we include a status field ('CREATED', 'SENT', 'SENT_FAILED', 'READ') into letter.entity.ts? I guess we can derive the value from the Notification's providerID and letter's firstReadAt? What are your thoughts on this? @AikoPath @gweiying


Yea, the status field was also something we had discussed. I'm okay with that to indicate whether a message has been published (to Twilio).
Not sure about the usage of `firstReadAt` to change the notification state. It feels like we are mixing data in that case. But a potentially following integration with the notification provider to get this info seems feasible.

backend/src/notifications/clients/twilio.service.ts
```diff
@@ -0,0 +1,60 @@
...
+            )
+            return Promise.resolve({
+              sid: `DEV_MODE_NO_SMS_SENT_${randomUUID().substring(0, 13)}`,
+            })
+          },
+        },
+      }
+    }
+
+    return twilio(accountSid, authToken)
```
thanhdatle, line 37: Should twilio service be a static instance and get initialised once?
The sendMessage function is called in a loop
`templatedMessages.map(async ({ message, letterId }, index) => {
        const recipient = recipients[index]
        const standardizedRecipient = this.standardizePhoneNumber(recipient)
        let messageSid
        try {
          messageSid = await this.smsClient.sendMessage(`
if the loops has 100 items, there will be 100 twilio clients getting generated. I am not 100% sure if this is recommended practice from Twilio.
`const client = this.getClient({
      logOnly: !this.isInSmsAllowList(recipient),
    })`

gweiying: That's a good point, I think that was my earlier intention as well, and I only instantiated it once. However with the allowlist, I couldn't figure out a way to nicely share the log-only implementation, so I ended up setting up a client multiple times. 
```
private getClient() {
    const accountSid = this.config.get('twilioApi.accountSid')
    const authToken = this.config.get('twilioApi.authToken')
    const messagingServiceSid = this.config.get('twilioApi.messagingServiceSid')

    if (accountSid && authToken && messagingServiceSid) {
      return twilio(accountSid, authToken)
    }

    return {
      messages: {
        create: ({ body, to }: { body: string; to: string }) => {
          this.logger.info(
            `Development mode: No SMS sent. Logging SMS: ${body} sent to ${to} `,
          )
          return Promise.resolve({
            sid: `DEV_MODE_NO_SMS_SENT_${generatePublicId().substring(0, 13)}`,
          })
        },
      },
    }
  }
  private client = this.getClient()

  async sendMessage(message: string, recipient: string): Promise<string> {
    // TODO: Add allow-listed phone numbers for dev testing
    this.logger.info(`Sending sms to ${recipient}`)
    const response = await this.client.messages.create({
      messagingServiceSid: this.config.get('twilioApi.messagingServiceSid'),
      body: message,
      to: recipient,
    })

    return response.sid
  }
```
Lemme think on it further!

backend/src/database/entities/notification.entity.ts
gweiying: Can also be email, or NRIC in the future if we implement other types of notification!

backend/src/notifications/notifications.service.ts
gweiying: Actually left it generic for now as we might extend to other kinds of recipient contact details e.g. email and NRIC! Do you think it might make sense to keep is as is for now?

backend/src/notifications/notifications.service.ts
gweiying: Actually left it generic for now as we might extend to other kinds of recipient contact details e.g. email and NRIC! Do you think it might make sense to keep is as is for now?

backend/src/database/migrations/1683596309672-update-letter-public-id-field.ts
```diff
@@ -24,8 +24,6 @@ export class updateLetterPublicIdField1683596309672
       `ALTER TABLE "letters" DROP CONSTRAINT "UQ_cfe06d7f8cda8902e25c03c0c70"`,
     )
     await queryRunner.query(`ALTER TABLE "letters" DROP COLUMN "publicId"`)
-    await queryRunner.query(
-      `ALTER TABLE "letters" ADD "publicId" text NOT NULL`,
-    )
+    await queryRunner.query(`ALTER TABLE "letters" ADD "publicId" text`)
```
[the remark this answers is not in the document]
gweiying: Yep, I had issues with reverting my migrations because of this!

backend/src/letters/letters-validation.service.spec.ts
gweiying: good idea, [11aaaa8](https://github.com/opengovsg/letters/pull/166/commits/11aaaa81ec02b9e6c80efbca2f6ed555d9c5fec1)

backend/src/types/notification.ts
gweiying: good point, [571a518](https://github.com/opengovsg/letters/pull/166/commits/571a518206e3c397bfec78687e36530c5e42db6d)

backend/src/notifications/clients/twilio.service.ts
AikoPath: @thanhdatle 
Excellent catch!

backend/src/notifications/clients/twilio.client.ts
```diff
@@ -13,47 +13,56 @@ export class TwilioClient {
...
+
+    let response
+    if (logOnly) {
+      response = await this.getLogOnlyClient().messages.create({
+        body: message,
+        to: recipient,
+      })
+    } else {
+      response = await this.client.messages.create({
+        messagingServiceSid: this.config.get('twilioApi.messagingServiceSid'),
```
thanhdatle, line 61: should this be a private field as well?

gweiying: I'm not entirely convinced that this is much cleaner since we still need to reference the variables but now also have the additional step of initializing it in the constructor, but lemme know what you think? Ambivalent either way [fa1ec66](https://github.com/opengovsg/letters/pull/166/commits/fa1ec66aa932bba35e07ec2f0502dec36d68b5de)

thanhdatle: makes sense, let's standardise this in another PR.

thanhdatle (approved):
LGTM! Thanks for working on this @AikoPath @gweiying

PR #167, opened 2023-06-30
fix: csv uploads error UI

PikkaPikkachu:
## Context

We are trying to improve the design for CSV uploads. In this PR specifically we will group the errors for each row together, so that's it's easier for users to process these errors. 

Closes this notion [task](https://www.notion.so/opengov/Eng-Error-handling-for-CSV-uploads-47eb937d69274a03b4ae551446d16f4c).

## Approach

Fixed the design. Did some front-end processing to group errors from the same row together using `reduce`.
Since we don't expect a ton of errors anyways (the scale of errors would be limited by the number of letters generated in one go), front-end processing would still make sense and should not be too heavy.

**Features**:
- Group related row errors together

## Before & After Screenshots

**BEFORE**:

Screenshot 2023-06-30 at 1 57 44 PM

**AFTER**:

https://github.com/opengovsg/letters/assets/25703976/60edb725-85b1-47f3-88e7-2f70a41eca9c

### DIFF
--- a/frontend/src/features/issue/BulkIssueDrawer/UploadCsvErrorsTable.tsx
+++ b/frontend/src/features/issue/BulkIssueDrawer/UploadCsvErrorsTable.tsx
@@ -42,6 +42,43 @@ export const UploadCsvErrorsTable = ({
     )
   }
 
+  const errorsByRow = uploadCsvErrors.reduce(
+    (errorGroup: Array<BulkLetterValidationResultError[]>, item) => {
+      if (errorGroup[item.id]) {
+        errorGroup[item.id].push(item)
+      } else {
+        errorGroup[item.id] = [item]
+      }
+      return errorGroup
+    },
+    new Array<BulkLetterValidationResultError[]>(uploadCsvErrors.length),
+  )
+
+  const errorTableRows = errorsByRow.map((errorData, index) => {
+    return (
+      <Tr key={JSON.stringify(index)}>
+        {/* We add 2 because the first row with data in the CSV is row 2, even though it has index 0 */}
+        <Td>{index + 2}</Td>
+        <Td>
+          {errorData.map((error, index) => {
+            return (
+              <Text key={index}>
+                {/* Add 1 because index starts from 0 actually */}
+                {index + 1}
+                {'. '}
+                {error.param}{' '}
+                {error.message ===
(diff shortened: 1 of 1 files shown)


frontend/src/features/issue/BulkIssueDrawer/UploadCsvErrorsTable.tsx
```diff
@@ -42,6 +42,43 @@ export const UploadCsvErrorsTable = ({
     )
   }
 
+  const processedErrors = uploadCsvErrors.reduce(
```
AikoPath, line 45: Based on what the function is doing I'd recommend to name it something along the lines of `errorsGroupdByRow` or `errorsByRow`.

frontend/src/features/issue/BulkIssueDrawer/UploadCsvErrorsTable.tsx
```diff
@@ -42,6 +42,43 @@ export const UploadCsvErrorsTable = ({
...
+        errorGroup[item.id].push(item)
+      } else {
+        errorGroup[item.id] = [item]
+      }
+      return errorGroup
+    },
+    new Array<BulkLetterValidationResultError[]>(uploadCsvErrors.length),
+  )
+
+  const errorTableData = processedErrors.map((errorData, index) => {
```
AikoPath, line 57: This looks like it is mapping the raw data into actual table rows and returning them.
We could call it something like `errorTableRows`.

AikoPath (changes requested):
Only two tiny naming suggestions. Otherwise lgtm!

FYI:
Your "after" video in the PR description briefly shows your slack window.
Given the open source nature of OGP it might be good to try to not include personal or internal information in our repos.

frontend/src/features/issue/BulkIssueDrawer/UploadCsvErrorsTable.tsx
PikkaPikkachu: Sounds good :)

PikkaPikkachu:
> Given the open source nature of OGP it might be good to try to not include personal or internal information in our repos.

Thanks! I removed the previous test video!

AikoPath (approved):

PR #170, opened 2023-07-04
fix: hide download button

PikkaPikkachu:
## Context

We are hiding the `Download as .PDF` button for now, since users can still use browser tools for printing the letters, moreover, the pagination of letters still remains a problem. 

Closes [notion task](https://www.notion.so/opengov/Eng-Hide-download-pdf-button-9ffd0cee684444c69d87adaa0645a9f9)

## Approach

Commented out code for the `Download as .PDF` button

**Bug Fixes**:

- We are hiding the "Download as .PDF" button 

## Before & After Screenshots

**BEFORE**:
Screenshot 2023-07-05 at 1 48 48 AM

**AFTER**:

https://github.com/opengovsg/letters/assets/25703976/db268808-333a-4de3-a64b-60386446abe4

### DIFF
--- a/frontend/src/features/public/LetterPublicPage.tsx
+++ b/frontend/src/features/public/LetterPublicPage.tsx
@@ -1,11 +1,11 @@
-import { Button, VStack } from '@chakra-ui/react'
+import { VStack } from '@chakra-ui/react'
 import { FormEvent, useEffect, useState } from 'react'
 import { Helmet } from 'react-helmet'
 import { Navigate } from 'react-router-dom'
 
 import { routes } from '~constants/routes'
 import { LetterViewer } from '~features/editor/components/LetterViewer'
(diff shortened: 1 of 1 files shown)


AikoPath (commented):
I believe just commenting out this bit results in linting errors as the `handleDownload` function is still declared but not used.

Commenting out that function (and its dependencies) as well would solve this issue.
However, there is the point to be made the solution for this task is to use source code version control instead of commenting out. After all, that is part of the purpose of version control.

PikkaPikkachu:
> Commenting out that function (and its dependencies) as well would solve this issue.
However, there is the point to be made the solution for this task is to use source code version control instead of commenting out. After all, that is part of the purpose of version control.

That makes sense, I wasn't super sure on why deleting would be different from hiding in that case haha. Therefore, just commented out.

gweiying (approved):
lgtm!

PR #171, opened 2023-07-04
chore: [citizen] password instructions

PikkaPikkachu:
## Context
In the previous, we added password instructions flow for the officers, in this PR we are going to be adding the flow for password instructions for citizens and how the password instructions would be shown to them before they actually open and access the letter.
Closes [insert issue #]

## Approach

- Passing the password instructions as part of the JSON error we pass, when password is required to block the letter.

**Features**:

- Show password instructions entered by the officers while generating the letters to the citizen when they access the letter. 

## Before & After Screenshots

**BEFORE**:
Enter password screen before: 
Screenshot 2023-07-05 at 1 30 09 PM

**AFTER**:
E2E flow tested here: 

https://github.com/opengovsg/letters/assets/25703976/8989b8dd-3ce6-47f0-85a9-ada449f1f986

When no password instructions are provided:
Screenshot 2023-07-06 at 6 38 55 PM

### DIFF
--- a/backend/src/batches/batches.service.ts
+++ b/backend/src/batches/batches.service.ts
@@ -1,4 +1,4 @@
-import { Injectable } from '@nestjs/common'
+import { Injectable, NotFoundException } from '@nestjs/common'
 import { InjectRepository } from '@nestjs/typeorm'
 import { EntityManager, Repository } from 'typeorm'
 
@@ -22,6 +22,15 @@ export class BatchesService {
   ): Promise<Batch> {
     const batch = this.repository.create(createBatchDto)
     return await entityManager.save(batch)
+  }
+
+  async findOneByBatchId(id: number): Promise<Batch> {
+    const batch = await this.repository.findOneBy({
+      id,
+    })
+    if (!batch) throw new NotFoundException('Batch not found')
+
+    return batch
   }
 
   findAll() {
--- a/backend/src/letters/letters.service.ts
+++ b/backend/src/letters/letters.service.ts
@@ -7,6 +7,7 @@ import {
 import { InjectRepository } from '@nestjs/typeorm'
 import { DataSource, EntityManager, Repository } from 'typeorm'
 
+import { PASSWORD_ERROR_MESSAGE } from '~shared/constants/letters'
 import { CreateBatchDto } from '~shared/dtos/batches.dto'
 import {
   CreateBulkLetterDto,
@@ -147,8 +148,18 @@ export class LettersService {
     if (!letter) throw new NotFoundException('Letter not found')
 
     if (letter.isPasswordProtected) {
+      // this is the case, where we show the screen for entering the password
+      // and the password instructions. So here we pass the password instructions
+      // to be displayed on frontend
       if (!password) {
-        throw new UnauthorizedException('No Password provided')
+        const letterBatch = await this.batchesService.findOneByBatchId(
+          letter.batchId,
+        )
+        throw new UnauthorizedException({
+          statusCode: 401,
+          message: PASSWORD_ERROR_MESSAGE,
+          passwordInstructions: letterBatch.passwordInstructions,
+        })
       }
 
       return this.lettersEncryptionService.decryptLetter(letter, password)
--- a/frontend/src/features/public/components/PasswordProtectedView.tsx
+++ b/frontend/src/features/public/components/PasswordProtectedView.tsx
@@ -1,4 +1,6 @@
 import {
+  Alert,
+  AlertIcon,
   Box,
   Button,
   FormControl,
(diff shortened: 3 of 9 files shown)


frontend/src/types/PasswordResponseError.ts
```diff
@@ -0,0 +1,10 @@
+export type PasswordResponseErrorJson = {
```
gweiying, line 1: nit here, would prefer if we kept our base type `ResponseErrorJson` and used `PasswordResponseErrorJson` to extend the type, as `ResponseErrorJson` can be used for all error status codes from our backend. For example, we can do something like this:

```
export type PasswordResponseErrorJson = ResponseErrorJson & {
  statusCode: number
}
``` 

Will leave you to decide how the files can be best organized.

frontend/src/features/public/components/PasswordProtectedView.tsx
```diff
@@ -70,11 +75,17 @@ export const PasswordProtectedView = ({
...
               mb="2.5rem"
             >
-              <FormLabel fontWeight={700}>Unlock Letter</FormLabel>
+              <FormLabel fontWeight={700}>
+                Instructions to Unlock Letter
+              </FormLabel>
+              <Alert status="info" style={{ margin: '10px 0px 10px' }}>
+                <AlertIcon />
+                <Text fontSize="sm">{passwordInstructions}</Text>
+              </Alert>
```
gweiying, line 84: nit, might want to do a conditional rendering for this chunk, as the Password Instructions are optional

backend/src/letters/letters.service.ts
```diff
@@ -147,8 +147,18 @@ export class LettersService {
...
+      // and the password instructions. So here we pass the password instructions
+      // to be displayed on frontend
       if (!password) {
-        throw new UnauthorizedException('No Password provided')
+        const letterBatch = await this.batchesService.findOneByBatchId(
+          letter.batchId,
+        )
+        throw new UnauthorizedException({
+          statusCode: 401,
+          message: 'Password not provided',
```
gweiying, line 159: Note that the message `'No Password provided'` is hardcoded on the frontend such that if the error message matches `'No Password provided'`, no error is shown to the user. This is to improve the user experience such that we're not throwing an error even before they try to unlock the letter.

Would suggest to move the message to the shared folder so that it doesn't break accidentally, like with this small wording change above. 
 
Screenshot 2023-07-05 at 4 52 09 PM

frontend/src/features/public/components/PasswordProtectedView.tsx
gweiying: Screenshot 2023-07-05 at 5 00 06 PM

frontend/src/features/public/components/PasswordProtectedView.tsx
```diff
@@ -70,11 +75,17 @@ export const PasswordProtectedView = ({
...
               <InputGroup>
                 <Input
                   type={showPassword ? 'text' : 'password'}
-                  placeholder="Enter password"
+                  placeholder="Start typing password"
```
gweiying, line 88: small suggestion, non-blocking but I did prefer the earlier text. is there a specific reason why we changed the copy here? will defer to any other discussions that you might have had with Pranu, Clement, Hena

gweiying (commented):
left some small comments inline! pretty minor changes, thanks for the quick turnaround

frontend/src/features/public/components/PasswordProtectedView.tsx
PikkaPikkachu: Hmm I saw the figma on task, and that's the copy it had, so made the relevant changes :)
I think we can keep the text that was there before too!

frontend/src/types/PasswordResponseError.ts
PikkaPikkachu: I think this makes sense! Mainly, thinking was that this error type is not being used for other use cases, but makes sense to keep it extensible for future :)

PikkaPikkachu:
Thanks @gweiying for the review! I have addressed the feedback and updates the test plan!

backend/src/batches/batches.service.ts
```diff
@@ -24,6 +24,17 @@ export class BatchesService {
     return await entityManager.save(batch)
   }
 
+  async findOneByBatchId(id: number): Promise<Batch> {
+    const batch = await this.repository.findOne({
+      where: {
+        id: id,
+      },
+    })
```
gweiying, line 32: small suggestion, non-blocking. this can be simplified to:
`const batch = await this.repository.findOneBy({ id })`

gweiying (approved):
left a small suggestion but non-blocking!

apart from that looks good to me, thanks

backend/src/batches/batches.service.ts
PikkaPikkachu: Thank you! Didn't know this ! :)

PR #175, opened 2023-07-06
fix: letter id generation

PikkaPikkachu:
## Context

This PR fixes the letter ID generation algo for generating a more "readable" letter ID. 

Closes this [task](https://www.notion.so/opengov/Eng-Update-letters-UUID-25a9255bb22f4b62b8adb7e8d84a4b91).

## Approach

1. Make letter ID length `size = GROUPS x BLOCK_SIZE`
2. We first generate the letter ID
3. Then insert the dividers in between using regex expression matching logic. 

## Before & After Screenshots

**BEFORE && AFTER**:
The screenshot below shows the newly generated letter ID vs the old letter ID [in the second row]
Screenshot 2023-07-06 at 4 12 01 PM

End to end testing video:

https://github.com/opengovsg/letters/assets/25703976/902da2f6-b2c9-4dfc-aec8-79dded042876

### DIFF
--- a/backend/src/core/utils/generate-public-id.spec.ts
+++ b/backend/src/core/utils/generate-public-id.spec.ts
@@ -12,12 +12,31 @@ describe('generatePublicId', () => {
     const mockGenerator = jest
       .fn()
       .mockReturnValueOnce('api')
-      .mockReturnValue('abcdefghi')
+      .mockReturnValue('abcdeabcdeabcdeabcde')
 
     const id = generatePublicId(mockGenerator)
 
     expect(mockGenerator).toHaveBeenCalledTimes(2)
     expect(PROTECTED_NAMESPACES).not.toContain(id)
-    expect(id).toEqual('abcdefghi')
+    expect(id).toEqual('abcde-abcde-abcde-abcde')
+  })
+
+  test('should throw error if ID string is not divisible by block size', () => {
+    const mockGenerator = jest
+      .fn()
+      .mockReturnValueOnce('api')
+      .mockReturnValue('abcdefgj')
+
+    expect(() => {
+      generatePublicId(mockGenerator)
+    }).toThrow()
+  })
+
+  test('should correctly insert delimeter in the id string generated', () => {
+    const mockGenerator = jest.fn().mockReturnValueOnce('23432dfdsf')
+
+    const id = generatePublicId(mockGenerator)
+
+    expect(id).toEqual('23432-dfdsf')
   })
 })
--- a/backend/src/core/utils/generate-public-id.spec.ts
+++ b/backend/src/core/utils/generate-public-id.spec.ts
@@ -11,14 +11,14 @@ describe('generatePublicId', () => {
   test('should regenerate ID if it matches a protected string', () => {
     const mockGenerator = jest
       .fn()
-      .mockReturnValueOnce('wheredream')
+      .mockReturnValueOnce('apicd')
       .mockReturnValue('abcdeabcdeabcdeabcde')
 
     const id = generatePublicId(mockGenerator)
 
-    expect(mockGenerator).toHaveBeenCalledTimes(2)
+    expect(mockGenerator).toHaveBeenCalledTimes(1)
     expect(PROTECTED_NAMESPACES).not.toContain(id)
-    expect(id).toEqual('abcde-abcde-abcde-abcde')
+    expect(id).toEqual('apicd')
   })
 
   test('should throw error if ID string is not divisible by block size', () => {
--- a/backend/src/core/utils/generate-public-id.spec.ts
+++ b/backend/src/core/utils/generate-public-id.spec.ts
@@ -11,7 +11,7 @@ describe('generatePublicId', () => {
   test('should regenerate ID if it matches a protected string', () => {
     const mockGenerator = jest
       .fn()
-      .mockReturnValueOnce('api')
+      .mockReturnValueOnce('wheredream')
       .mockReturnValue('abcdeabcdeabcdeabcde')
 
     const id = generatePublicId(mockGenerator)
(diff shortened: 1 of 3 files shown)


AikoPath (commented):
Do we expect the id format to change again?
If not we can just do `${a.slice(0,5)}-${a.slice(5,10)}-${a.slice(10,15)}-${a.slice(15,20)}`, which would be significantly easier to understand.

backend/src/core/utils/generate-public-id.ts
```diff
@@ -22,19 +24,24 @@ const customRandom = (
...
+const customAlphabet = (alphabet: string, blockSize: number, size: number) =>
+  customRandom(alphabet, size, blockSize, random)
 
-const ALPHABET =
-  '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz'
-const ID_LENGTH = 32
+const ALPHABET = '0123456789abcdefghijklmnopqrstuvwxyz'
+const BLOCK_SIZE = 5
+const GROUPS = 4
+const ID_LENGTH = BLOCK_SIZE * GROUPS + GROUPS - 1
```
gweiying, line 41: Small nit, just thinking that this is somewhat hard to read and abit difficult to maintain. Might be related to what Aiko has already mentioned.

What if we separated the function for generating the public id into 2 functions, one for generating a random string of length ID_LENGTH e.g. `generateRandomStr(length=ID_LENGTH)`, and another for splitting the id generated into blocks `splitStrIntoBlocks(groups = GROUPS, blockSize = BLOCK_SIZE)`? Might improve readability while avoiding hardcoded stuff.

PikkaPikkachu: Sounds good, we can do that!

PikkaPikkachu: Hmm on a second thought, in that case, we would still be defining `ID_LENGTH = BLOCK_SIZE * GROUPS` right ? But just not the `GROUPS -1` part ?

gweiying: Ah yes correct

backend/src/core/utils/generate-public-id.ts
```diff
@@ -26,12 +26,31 @@ const customRandom = (
...
+  // throws error if the stirng is not divisible by blockSize
+  if (stringToSplit.length % blockSize != 0) {
+    throw new Error('The string cannot be split into equal block sizes.')
+  }
+  // generate regex pattern for recurrently matching a string length of blockSize
+  const matchBlockSizeRegex = new RegExp(`(.{${blockSize}})`, 'g')
+
+  // now, simply match with the above generated regex and insert dividers
+  // also remove the last character to remove the extra added '-' at the end using .slice(0, -1)
+  return stringToSplit.replace(matchBlockSizeRegex, '$1' + divider).slice(0, -1)
```
gweiying, line 44: Something like this might also be abit easier to maintain and need less inline documentation, I was confused reading the `.slice(0, -1)` part.

```
const stringBlockArr = stringToSplit.match(/.{blockSize}/g)
return stringBlockArr.join('-')
```

PikkaPikkachu: Ahh I tried this! 
Apparently, there's one small problem with using `.match` which either returns a `null|RegexArray`
So we will have do a separate null check and potentially throw an error in letter ID generation if `stringBlockArr` is null — if you think that's still better then we can do that.

shared/src/constants/protected-namespaces.ts
```diff
@@ -10,4 +10,6 @@ export const PROTECTED_NAMESPACES = [
   'privacy',
   'guide',
   'not-found',
+  // adding one for testing purposes
+  'where-dream',
```
gweiying, line 14: nit, not sure if this was left here accidentally?

backend/src/core/utils/generate-public-id.ts
gweiying: Hey! I realized there was an issue with using `match(/.{blockSize}/g)` that `blockSize` was not being recognized as a variable. 

What do you think of the following suggestion that combines our implementations?
```
const splitStrIntoBlocks = (
  stringToSplit: string,
  blockSize: number,
  divider = '-',
) => {
  if (stringToSplit.length % blockSize != 0) {
    throw new Error('The string cannot be split into equal block sizes.')
  }
  const blockSizeRegex = new RegExp(`(.{${blockSize}})`, 'g')
  const stringBlockArr = stringToSplit.match(blockSizeRegex) || []
  return stringBlockArr.join(divider)
}
```

I've removed some of the inline documentation because I think the documentation is abit difficult to read and the function+function naming should be clear enough to show what the function is doing. I've also modified your replace function to a simple join, which should also remove the need to explain what `$1` is and the slicing.

gweiying (commented):
overall looks ok to me! I've just left a comment inline on some tips to improve the readability of the code, we can go over it if it's unclear!

backend/src/core/utils/generate-public-id.ts
PikkaPikkachu: > const stringBlockArr = stringToSplit.match(blockSizeRegex) || []

Hmm, I was just a teeny bit skeptical of this statement, because if we are allowing this, then it potentially means, the letter ID generated is empty and we would go ahead and generate letters, with a "" letter ID ?

If that's not really a concern, then this sounds good to me:)

gweiying (approved):

backend/src/core/utils/generate-public-id.spec.ts
```diff
@@ -11,13 +11,32 @@ describe('generatePublicId', () => {
   test('should regenerate ID if it matches a protected string', () => {
     const mockGenerator = jest
       .fn()
-      .mockReturnValueOnce('api')
-      .mockReturnValue('abcdefghi')
+      .mockReturnValueOnce('wheredream')
```
gweiying, line 14: quick note that this needs to be rolled back!

PR #176, opened 2023-07-06
chore: add password validation

PikkaPikkachu:
## Context

Closes this [task](https://www.notion.so/opengov/Eng-Password-validation-5083a069d9d245fdb32d5e371eba481c)

## Approach
1. Added two more error types for short password and alphanumeric password
2. Based on the condition not met, we show the same error to the user 

One thing I was not too sure about was whether we should be grouping these errors, but then it might not be super obvious what is the exact error. 

**Features**:
- Add password validation for when the password entered by the officer is less than 8 characters 
- Add password validation for when the password entered by the officer is not alphanumeric

## Before & After Screenshots

**BEFORE**:
Errors before making the changes for password validation:
Screenshot 2023-07-06 at 5 17 44 PM

**AFTER**:
Tested on various test cases:
Screenshot 2023-07-06 at 5 15 59 PM

This is what the uploaded CSV looks like:
Screenshot 2023-07-05 at 5 09 02 PM

### DIFF
--- a/backend/src/letters/letters-validation.service.ts
+++ b/backend/src/letters/letters-validation.service.ts
@@ -1,6 +1,10 @@
 import { Injectable } from '@nestjs/common'
 
-import { BULK_MAX_ROW_LENGTH } from '~shared/constants/letters'
+import {
+  ALPHANUMERIC_PASSWORD_REGEX,
+  BULK_MAX_ROW_LENGTH,
+  MIN_PASSWORD_LENGTH,
+} from '~shared/constants/letters'
 import {
   BulkLetterValidationResultDto,
   BulkLetterValidationResultError,
@@ -40,7 +44,7 @@ export class LettersValidationService {
     )
 
     if (passwords) {
-      errors.push(...this.validatePasswordsArePresent(passwords))
+      errors.push(...this.validatePasswordsArePresentAndValid(passwords))
     }
 
     if (errors.length === 0)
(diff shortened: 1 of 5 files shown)


backend/src/letters/letters-validation.service.ts
```diff
@@ -88,17 +92,36 @@ export class LettersValidationService {
...
+              param: 'Password',
+              message: BulkLetterValidationResultErrorMessage.SHORT_PASSWORD,
+            }
+          : {
+              id: password.initialIndex,
+              param: 'Password',
+              message:
+                BulkLetterValidationResultErrorMessage.PASSWORD_NOT_ALPHANUMERIC,
+            },
+      )
```
AikoPath, line 124: - I am not sure if the `filter` + `map` is a good option here, as it requires us check the same conditions twice. This could be done with a `forEach`.
- The chained ternary `?` operator might also be a little less readable than a switch or if clauses.
- As only the `message` changes, the branching could just be on that field.

What do you think of something like this?
(I have not tested this :) )
```
const errors: BulkLetterValidationResultError[] = []

passwords.forEach((password, index) => {
  let errorMessage: BulkLetterValidationResultErrorMessage | null = null
  if (password === '') {
    errorMessage = BulkLetterValidationResultErrorMessage.MISSING_PARAM
  } else if (password.length < MIN_PASSWORD_LENGTH) {
    errorMessage = BulkLetterValidationResultErrorMessage.SHORT_PASSWORD
  } else if (!password.match(ALPHANUMERIC_PASSWORD_REGEX)) {
    errorMessage = BulkLetterValidationResultErrorMessage.PASSWORD_NOT_ALPHANUMERIC
  }
  if (errorMessage) {
    errors.push({
      id: index,
      param: 'Password',
      message: errorMessage,
    })
  }
})
return errors
```

gweiying:
Raising a broader discussion point, so commenting inline here. 
- Is there a particular reason we decided to go with alphanumeric only instead of supporting any ASCII characters? Since it's a security mechanism and not something designed for readability, I see less value in restricting it to alphanumeric. Also symbols such as dashes, / or * # might be common in passwords especially for birthdays etc.

frontend/src/features/issue/BulkIssueDrawer/UploadCsvErrorsTable.tsx
```diff
@@ -42,6 +42,27 @@ export const UploadCsvErrorsTable = ({
...
+    if (
+      errorMessage ===
+      BulkLetterValidationResultErrorMessage.PASSWORD_NOT_ALPHANUMERIC
+    ) {
+      return ' must be alphanumeric.'
+    }
+    if (
+      errorMessage === BulkLetterValidationResultErrorMessage.SHORT_PASSWORD
+    ) {
+      return ' must contains a minimum of 8 characters.'
```
AikoPath, line 61: Should we just define the errors in the backend instead of having a 1 to 1 match from a string to another string in the front end?
With the SMS sending coming as well this mapping would grow quite big.

AikoPath:
> Raising a broader discussion point, so commenting inline here.
> 
> * Is there a particular reason we decided to go with alphanumeric only instead of supporting any ASCII characters? Since it's a security mechanism and not something designed for readability, I see less value in restricting it to alphanumeric. Also symbols such as dashes, / or * # might be common in passwords especially for birthdays etc.

+ 100

frontend/src/features/issue/BulkIssueDrawer/UploadCsvErrorsTable.tsx
```diff
@@ -42,6 +42,27 @@ export const UploadCsvErrorsTable = ({
     )
   }
 
+  const getErrorMessage = (errorMessage: string): string => {
+    console.log('this is the error messsage', errorMessage)
```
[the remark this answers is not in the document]
PikkaPikkachu: Oops!

backend/src/letters/letters-validation.service.ts
PikkaPikkachu: Thanks for the suggestion, sounds good! :)

PR #182, opened 2023-07-11
Feat/a4 resize

gweiying:
## Context
Responsive letters behave poorly on small screensizes as we are unable to control how page breaks and html formatting (e.g. tables, boxes) resize on small screens.

To resolve this, we make our letters non-responsive and retain the A4 fixed ratio, so that all letters to look identical on all screensizes. 

## Approach

- Adds a `useTransformScale` hook to detect screen size and resize `LetterViewer` when rendering `LetterPublicPage`. On screensizes smaller than A4, letters are rendered at 90% of screen width. On screensizes larger than A4, letters are rendered at A4 size
- For previews of letters in Drawers, letters are rendered at 90% of `DRAWER_WIDTH_IN_PIXELS`. Note that `DRAWER_WIDTH_IN_PIXELS` was hardcoded as it was difficult to extract the width of the parent component in React, and I didn't think it was worth additional engineering effort to fix as the revamped designs in Figma all do not seem to include previews in Drawers. 

## Before & After Screenshots

**BEFORE**:
Screen Recording 2023-07-11 at 5 20 40 PM

**AFTER**:
Screen Recording 2023-07-11 at 5 20 09 PM

Letters that contain multiple pages
Screen Recording 2023-07-11 at 3 40 31 PM

Screen Recording 2023-07-11 at 4 51 50 PM

## Bugs
Screen Recording 2023-07-11 at 4 53 52 PM
Note that on small screen sizes, there's an awkward gap between the end of the letter and the start of the footer. That's because we're only scaling the letter, not the entire page, and the scaling is done on via CSS transform (non-responsively). This means the page is unable to detect that the letter height has changed, so the footer stays the same distance away from the top of the page as if it was viewed on normal screen. There does not seem to be an easy way to resolve this for now unless we know a fixed height for the Letter, (then we could just scale the box height to be `(LetterViewer.height + padding)* transformScale`. For now, it seems like a small issue as users do not have any other functionality to click on below this gap.

Edit: ^ This is fixed for the public view

### DIFF
--- a/frontend/src/features/issue/BulkIssueDrawer/PreviewTemplate.tsx
+++ b/frontend/src/features/issue/BulkIssueDrawer/PreviewTemplate.tsx
@@ -1,7 +1,9 @@
 import { Box } from '@chakra-ui/react'
 import { Button } from '@opengovsg/design-system-react'
 
+import { DRAWER_WIDTH_IN_PIXELS } from '~constants/theme'
 import { LetterViewer } from '~features/editor/components/LetterViewer'
+import { calculateTransformScale } from '~features/public/hooks/useTransformScale'
 
 import { useGetTemplateById, useTemplateId } from '../hooks/templates.hooks'
 
@@ -18,7 +20,14 @@ export const PreviewTemplate = ({
     <>
       <Box padding={8} flex="1" overflowY="auto" paddingBottom="6rem">
         {/* paddingBottom height to account for sticky button */}
-        <LetterViewer html={template?.html} isLoading={isTemplatesLoading} />
+        <LetterViewer
+          html={template?.html}
+          isLoading={isTemplatesLoading}
+          transformOrigin="top"
+          transform={`scale(${calculateTransformScale(
+            DRAWER_WIDTH_IN_PIXELS,
+          )})`} // TODO: should be able to avoid hardcoding by reading from parent drawer component
+        />
       </Box>
       <Box
         position="fixed"
--- a/frontend/src/features/public/LetterPublicPage.tsx
+++ b/frontend/src/features/public/LetterPublicPage.tsx
@@ -1,11 +1,10 @@
-import { Box, useDimensions, VStack } from '@chakra-ui/react'
-import { FormEvent, useEffect, useRef, useState } from 'react'
+import { Box, VStack } from '@chakra-ui/react'
+import { FormEvent, useEffect, useState } from 'react'
 import { Helmet } from 'react-helmet'
 import { Navigate } from 'react-router-dom'
 
 import { routes } from '~constants/routes'
 import { LetterViewer } from '~features/editor/components/LetterViewer'
-import { HEIGHT_A4 } from '~utils/htmlUtils'
 
 import { PasswordProtectedView } from './components/PasswordProtectedView'
 import {
--- a/frontend/src/features/public/LetterPublicPage.tsx
+++ b/frontend/src/features/public/LetterPublicPage.tsx
@@ -1,23 +1,25 @@
-import { VStack } from '@chakra-ui/react'
+import { Box, VStack } from '@chakra-ui/react'
 import { FormEvent, useEffect, useState } from 'react'
 import { Helmet } from 'react-helmet'
 import { Navigate } from 'react-router-dom'
 
 import { routes } from '~constants/routes'
 import { LetterViewer } from '~features/editor/components/LetterViewer'
-import { HEIGHT_A4, WIDTH_A4 } from '~utils/htmlUtils'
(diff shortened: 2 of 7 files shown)


frontend/src/features/public/LetterPublicPage.tsx
```diff
@@ -49,29 +51,28 @@ export const LetterPublicPage = (): JSX.Element => {
...
-        ) : (
-          <VStack padding={16} spacing={8} align={'center'}>
+      {isPasswordProtected && !letter ? (
+        <PasswordProtectedView
+          // eslint-disable-next-line @typescript-eslint/no-misused-promises
+          handleSubmit={handleSubmit}
+          error={error}
+          password={password}
+          setPassword={setPassword}
+          isLetterLoading={isLetterLoading}
```
PikkaPikkachu, line 61: Quick question — Is the removal of `passwordInstructions` here intentional?

frontend/src/features/issue/BulkIssueDrawer/PreviewTemplate.tsx
```diff
@@ -18,7 +20,14 @@ export const PreviewTemplate = ({
...
       <Box padding={8} flex="1" overflowY="auto" paddingBottom="6rem">
         {/* paddingBottom height to account for sticky button */}
-        <LetterViewer html={template?.html} isLoading={isTemplatesLoading} />
+        <LetterViewer
+          html={template?.html}
+          isLoading={isTemplatesLoading}
+          transformOrigin="top"
+          transform={`scale(${calculateTransformScale(
+            DRAWER_WIDTH_IN_PIXELS,
+          )})`} // TODO: should be able to avoid hardcoding by reading from parent drawer component
```
PikkaPikkachu, line 29: Just to confirm — are you planning to address the `TODO` within this PR only, or is this for later?

PikkaPikkachu:
Just a few teeny comments, overall logic LGTM!

AikoPath (commented):
Overall this looks good, just some small things to clean up.

Given that our goal for this story was to support mobile devices and our overall goal since the last prod review was to polish the app I had a look at the issue we are having due to CSS transform not affecting the document flow and might have found a solution.
Let catch up offline and see if it's possible to include that.

frontend/src/features/issue/BulkIssueDrawer/PreviewTemplate.tsx
gweiying: Thanks for asking, it's not something I plan to resolve within this PR.

frontend/src/features/public/LetterPublicPage.tsx
```diff
@@ -1,23 +1,25 @@
...
-import { FormEvent, useEffect, useState } from 'react'
+import { Box, useDimensions, VStack } from '@chakra-ui/react'
+import { FormEvent, useEffect, useRef, useState } from 'react'
 import { Helmet } from 'react-helmet'
 import { Navigate } from 'react-router-dom'
 
 import { routes } from '~constants/routes'
 import { LetterViewer } from '~features/editor/components/LetterViewer'
-import { HEIGHT_A4, WIDTH_A4 } from '~utils/htmlUtils'
+import { HEIGHT_A4 } from '~utils/htmlUtils'
```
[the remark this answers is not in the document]
gweiying: [0813fdd](https://github.com/opengovsg/letters/pull/182/commits/0813fddb05691204002e8e93d9448813fda359ab)

frontend/src/features/public/LetterPublicPage.tsx
gweiying: [3bbedd3](https://github.com/opengovsg/letters/pull/182/commits/3bbedd3e75d005a1d966b5263215a78d0ebeaa0f)

frontend/src/features/public/hooks/useTransformScale.ts
```diff
@@ -0,0 +1,27 @@
...
+    setTransformScale(calculateTransformScale(window.innerWidth))
+  }, [])
+
+  useEffect(() => {
+    window.addEventListener('resize', handleResize)
+
+    return () => {
+      window.removeEventListener('resize', handleResize)
+    }
+  }, [handleResize])
```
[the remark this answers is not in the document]
gweiying: I don't have any linting errors here :sob:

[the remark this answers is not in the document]
AikoPath: Oh yes, because `handleResize` is in the dependencies array. Without this our linting raises an error.
But, I wonder if it actually needs to be there?

AikoPath (approved):
Good refactoring! Makes a lot of sense to keep the stuff to get the height in the parent container 👍

PR #189, opened 2023-07-18
chore: add password validation

PikkaPikkachu:
## Context

NOTE: Will add tests, once the logic looks fine, there are a couple of changes due to which tests need modification. 

We are adding password validation for password protected letters. 
Closes this [task](https://www.notion.so/opengov/2ccadbea5c734aa09614b4ca8cc7177e?v=76cff9fca57b496f857c990e2639af2f&p=05631885c2a449e2bff9246921beab0b&pm=s)

## Approach
1. Added in a field for user displayed error message, which will be shown on the frontend and removed the frontend logic for deciding and hardcoding what error message should be shown. 
2. Added validation on the password string as required using REGEX comparisons. 

## Before & After Screenshots
The CSV file I uploaded both times:
Screenshot 2023-07-18 at 8 12 34 AM

**BEFORE**:

Screenshot 2023-07-18 at 8 11 42 AM

**AFTER**:
Screenshot 2023-07-18 at 7 38 36 AM

## Tests

TBA. Will add once the overall logic looks fine :)


backend/src/letters/letters-validation.service.ts
```diff
@@ -117,21 +123,47 @@ export class LettersValidationService {
...
+      } else if (!password.match(ALPHANUMERIC_REGEX)) {
+        errors.push({
+          id: index,
+          param: 'Password',
+          message: BulkLetterValidationResultErrorMessage.INVALID_PASSWORD,
+          displayedErrorMessage: ' should contain a mix of letters and numbers',
+        })
+      }
+    })
+    return errors
```
gweiying, line 166: This seems to conflict with the style of handling errors in https://github.com/opengovsg/letters/pull/193, but I do like it better! It's cleaner than doing the frontend check and parsing.

Can you and Keith combine the changes from #193 and this PR?

backend/src/letters/letters-validation.service.ts
```diff
@@ -117,21 +123,47 @@ export class LettersValidationService {
...
+          param: 'Password',
+          message: BulkLetterValidationResultErrorMessage.INVALID_PASSWORD,
+          displayedErrorMessage: ' contains non-ASCII characters',
+        })
+      } else if (!password.match(ALPHANUMERIC_REGEX)) {
+        errors.push({
+          id: index,
+          param: 'Password',
+          message: BulkLetterValidationResultErrorMessage.INVALID_PASSWORD,
+          displayedErrorMessage: ' should contain a mix of letters and numbers',
```
gweiying, line 162: Small suggestion, we can directly add the params into the `displayedErrorMessage` e.g. "Passwords should contain a mix of letters and numbers" so that we don't have to do the additional parsing on the frontend `{error.param} {' '} {error.displayedErrorMessage}`

backend/src/letters/letters-validation.service.ts
PikkaPikkachu: Yup planning to, I'll let @disKeith merge the error one first and build on top of it!

backend/src/letters/letters-validation.service.ts
gweiying, line 162: Other small nit is that I believe we decided to standardise on a single error copy for invalid passwords, which will simplify this check!

PikkaPikkachu: Hmm makes sense ! I'll address the change in the rebase/merge with Keith's PR.

PR #190, opened 2023-07-18
feat: sms issue download flow

disKeith:
## Context

This PR adds the download card as the last step to the SMS enabled bulk issue flow, and allows for user to see the phone numbers used to send the SMS notifications in the completed csv file.

## Approach

1. Added `phoneNumbers` state to store phone numbers from uploaded csv
2. Added an additional step to show `BulkIssueCompletionCard` when `CitizenNotificationMethod.SMS` is true
3. Added optional `phoneNumbers` parameter to `BulkIssueCompletionCard` and conditionally handle and return phone numbers to the completed csv for download
4. Removed SMS enabled issue flow's navigating back to dashboard with success toast after generate & sending letters at `BulkIssueSendLettersCard` and step to the `BulkIssueCompletionCard` instead

## Before & After Screenshots

**BEFORE**:

https://github.com/opengovsg/letters/assets/119388168/4b56c02a-105a-4055-b8e8-6b2781507939

Uploaded CSV:
image

**AFTER**:

https://github.com/opengovsg/letters/assets/119388168/cb628554-c4a1-40f8-8303-07c33303ba27

Uploaded CSV:
image

Completed CSV:
image


PikkaPikkachu:
Overall LGTM, I think we should merge with the current branch (and push to staging). Also I think we discussed we will rename the template header like so: `Settings` > `Upload file` > `Send & Download`/ `Download & Send`, so probably we can make that tiny change here as well :)

frontend/src/features/templates/BulkIssuePage.tsx
```diff
@@ -47,10 +51,22 @@ export const BulkIssuePage = (): JSX.Element => {
...
+                onSuccess={(letters: GetBulkLetterDto[]) => {
+                  setBulkLetters(letters)
+                }}
+                shouldDisplay={activeStep === 2}
+                goToNext={() => goToNext()}
+                goToPrevious={() => goToPrevious()}
+                setPhoneNumbers={setPhoneNumbers}
+              />
+              <BulkIssueCompletionCard
+                shouldDisplay={activeStep === 3}
```
PikkaPikkachu, line 65: out of curiosity, but we display the completion card as part of the same step i.e. `Send & download` is one step, so why do we need to increment the `activeStep` here actually ?

PR #198, opened 2023-07-24
Chore/fix encryption

YangXingHao96:
## Context

This PR is to address the issue of potential error in decryption of encrypted letterFields and issuedLetter. Previous decryption handles the cases of malformed utf8, empty results after decryption due to password error. However there is small chance that the text can be decrypted into utf8 context. Thus an anchorText is added to ensure that the message is correctly decrypted. For more details refer to the notion design doc: https://www.notion.so/opengov/20230719-Encryption-Decryption-Bug-Fix-0ff8890353a340f88dfc3e3f8a47a495

Closes [insert issue #]

## Approach

Additional column anchor text is added to letter table, perform database migration to add a field that is of type string and nullable, perform validation against default anchorText = OpenGovSg, which is saved as constant variable in repository. (should we modify this part to save it in database? like having an additional pre-encrypted anchor column?). If field is null, ignore and perform previous encryption, else, decrypt the content and validate against the default anchorText.

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:
[insert screenshot here]

**AFTER**:
[insert screenshot here]

## Tests

## Deploy Notes

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:

## Risks


backend/src/database/entities/letter.entity.ts
```diff
@@ -51,4 +51,7 @@ export class Letter {
 
   @Column({ type: 'timestamptz', nullable: true, default: null })
   firstReadAt: Date
+
+  @Column({ type: 'text', nullable: true })
+  anchorText: string
```
gweiying, line 56: small suggestion, it was abit confusing to follow the anchorText variable throughout the code since we're reusing the variable name `anchorText` for both the encrypted anchor text and the decrypted anchor text. 

Let's call this column `encryptedAnchorText` for clarity?

backend/src/letters/letters-encryption.service.ts
```diff
@@ -29,12 +32,21 @@ export class LettersEncryptionService {
       enc.Utf8,
     )
 
+    let anchorText: string
+    if (letter.anchorText === null) {
```
gweiying, line 36: Suggestion, let's add a short comment here to explain why anchorText might be null, something simple like `// we did not encrypt anchor text previously, so older password-protected letters do not contain the encryptedAnchorText column`

backend/src/letters/letters-encryption.service.ts
```diff
@@ -29,12 +32,21 @@ export class LettersEncryptionService {
       enc.Utf8,
     )
 
+    let anchorText: string
```
gweiying, line 35: Suggestion, let's call this variable `decryptedAnchorText` for clarity, related to the comment above!

backend/src/database/migrations/1689908223807-add-anchortext.ts
```diff
@@ -0,0 +1,13 @@
+import { MigrationInterface, QueryRunner } from 'typeorm'
+
+export class addAnchortext1689908223807 implements MigrationInterface {
+  name = 'addAnchortext1689908223807'
```
gweiying, line 4: tiny nit, mostly for future reference. would be nice to standardize on camelcase throughout the name, and also specify in the migration name which table the column is added to, for example `add-anchor-text-column-to-letters'

backend/src/letters/letters-encryption.service.spec.ts
```diff
@@ -67,5 +85,25 @@ describe('LettersEncryptionService', () => {
         )
       }).toThrow() // wrongly decrypted long cypher texts can cause 'Malformed UTF-8 data'
     })
+
```
gweiying, line 88: Small suggestion, what do you think of adding a test for backward compatibility, for when `anchorText` is null?

e2e/package.json
```diff
@@ -10,6 +10,7 @@
     "playwright": "^1.34.3"
   },
   "devDependencies": {
-    "@playwright/test": "^1.34.3"
+    "@playwright/test": "^1.34.3",
```
gweiying, line 13: Nit, this and the `e2e/package-lock.json` changes are outside of the scope of this PR. Do you mind rebasing your branch off develop so we can keep the changes isolated?

gweiying (approved):
Overall the PR looks good! 

Left minor comments about naming clarity and tests. Your documentation in Notion was quite comprehensive, so thanks for that. I left a small suggestion for improvement there!

One other small thing:
Your branch contains changes from your `chore/fix_login_test` branch. To keep the changes isolated, could you grab the relevant changes and branch off `develop`? This would be the regular convention for our branching strategy, so good practice to do it now even though the changes are quite small here.

backend/src/letters/letters-encryption.service.ts
YangXingHao96: added comments to indicate why it might be null

backend/src/letters/letters-encryption.service.ts
YangXingHao96: changed the variable naming

backend/src/letters/letters-encryption.service.spec.ts
YangXingHao96: added a new test, that has anchorText as null

e2e/package.json
YangXingHao96: removed all changes in e2e test

PR #199, opened 2023-07-24
Feat/sample sms new

gweiying:
## Context

We want to allow users to send sample SMSes to preview their letter as well as their SMS.

## Approach

The technical document for this is written on [Notion](https://www.notion.so/opengov/20230714-Send-sample-SMS-37bb90eb9f91444b8ee4991fd0daaa76).

## Before & After Screenshots

**AFTER**:
Screen Recording 2023-07-25 at 12 43 46 PM
RPReplay_Final1690260388

## Deployment notes
Sample letters are only created on template creation. For existing templates, we will need to create sample letters manually and add them to templates. This is already done on staging, but when we deploy this to production, we must add sample letters manually!


frontend/src/features/issue/components/modals/SendSampleSmsModal.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const handleSubmit = () => {
+    void mutateAsync({ templateId, phoneNumber })
+    onClose()
+    onSuccess()
+  }
+
+  return (
+    <Modal
+      size="lg"
+      blockScrollOnMount={false}
```
PikkaPikkachu, line 58: nit: this is probably not required ?

frontend/src/features/issue/components/BulkIssueSendLettersCard.tsx
```diff
@@ -91,9 +110,24 @@ export const BulkIssueSendLettersCard = ({
       <Text textStyle={'body-1'}>
         If you face issues viewing, please write to letters@open.gov.sg.
       </Text>
+      <Divider paddingTop="12px" />
+      <HStack
+        paddingTop="12px"
+        textStyle="subhead-1"
+        color="#005DEA"
```
PikkaPikkachu, line 117: hmm we don't have a theme color that we can use here?

backend/src/letters/letters.service.ts
```diff
@@ -223,6 +264,45 @@ export class LettersService {
...
+      },
+      {},
+    )
+    const sampleLetter = await this.singleCreate(userId, {
+      letterParamMap: sampleParamMap,
+      templateId: template.id,
+    })
+
+    await this.templatesService.update(template.id, { sampleLetter })
+    return
```
PikkaPikkachu, line 288: nit: do we need this explicit return here?

gweiying:
Replaced by #204

gweiying:
Replaced by #204

PR #200, opened 2023-07-25
fix: create template whitespaces

disKeith:
## Context

This PR addresses the issue of parsed html not retaining its original formatting after template creation, even though the formatting was done correctly on the template editor

## Approach

Current logic in `TemplatesParsingService` replaces all `&nbsp;` characters with white spaces to correctly extract valid fields in the html, the approach is to no longer do any character replacement on the parsed html, and handle the extraction of `validFields` using regex:

- Updated `ACCEPTED_TEMPLATE_FIELDS_REGEX` to handle trailing and leading instances for both white space and `&nbsp;` instances 

  - image
- Updated `getTemplateFields` to extract unique matched fields properly (inclusive of whitespace/`&nbsp;`) for subsequent parsing
- Added replacing `&nbsp;` with whitespaces back to `parseTemplateField` to properly trim fields
- Removed logic in `TemplatesParsingService` to replace `&nbsp;` with whitespaces


gweiying (approved):
lgtm!

shared/src/util/templates.ts
```diff
@@ -6,10 +6,14 @@ export const getTemplateFields = (html: string): string[] => {
   let match: RegExpExecArray | null
 
   while ((match = ACCEPTED_TEMPLATE_FIELDS_REGEX.exec(html)) !== null) {
-    if (!fields.includes(match[1])) fields.push(match[1])
+    if (!fields.includes(match[0].slice(2, -2)))
+      fields.push(match[0].slice(2, -2))
```
PikkaPikkachu, line 10: teeny suggestion: can we add some comments, on what this does ? Not too intuitive 😅

PikkaPikkachu (approved):
LGTM!

PR #204, opened 2023-07-27
Feat/sample sms refactored

gweiying:
## Context

We want to allow users to send sample SMSes to preview their letter as well as their SMS.

## Approach

The technical document and discussions for this is documented on [Notion](https://www.notion.so/opengov/20230714-Send-sample-SMS-37bb90eb9f91444b8ee4991fd0daaa76). A quick summary is provided below:

### Creating a sample letter
We considered two key approaches to trigger the generation of sample letters (other approaches were considered briefly and can be found in the Notion document above):

1. event-driven (where we fire an event after template creation and listen to the event to trigger the sample letter creation),  or
2. using the conventional approach where TemplatesService will call LettersService to create a sample letter.

The latter approach causes a module-level circular dependency between Templates and Letters even though no services are directly importing each other, because LettersController still uses TemplatesService. We will need to use forwardRef() so that the modules can compile at runtime

However, we decided that this is acceptable as it doesn't cause any serious runtime issues (unlike file-level circular dependencies), and preferable than maintaining the event-driven approach as the event-driven approach would be harder to trace and debug especially for new engineers without context on the project.

**Decision: Use forwardRef() to handle module-level circular dependencies between Templates and Letters.**

### Sending a test notification 
Currently our bulk letter generation and bulk notification are all triggered via the endpoint `POST /letters/bulks`. In a RESTful approach we would want our test notification to be triggered by a `POST /notifications/sample` endpoint. 

However, this risks introducing further circular dependencies between Letters, Notifications and Templates modules because Letters module would use Notifications module to trigger notifications, and Notifications module would use Letters module (via Templates modules) to find the sample letter. 

We discussed that the ideal approach would be to refactor the current codebase to expose a notification endpoint `POST /notifications/bulks`, and another for `POST /notifications/sample`. This would allow Notifications to be a standalone controller instead. 

**Decision: Expose a `POST /letters/sample` to send sample notifications from LettersController (same with how bulk notifications are currently sent). Add an open task to refactor the codebase so that all notifications will be triggered from the Notifications controller.**  

### Other changes
- Added backend validation on SMS sending to throw errors on non-SG numbers
  - This helps to validate the phone number for sample SMS sending 
  - For bulk generation, this should have no effect on users as phone numbers are already validated on CSV upload, however this preempts the move to a separate `POST /notifications/bulks` endpoint as the modules should not be aware of each other's logic i.e. Notifications module should do its own validation on phone numbers as it should not be aware that Letters module has already validated it
- Created a letters service `singleCreate` method to be used for creating the sample letter. 
  - If we plan to take in users' input and allow for single letter creation, we will need to add a `validateSingle` letter method. Currently there's only a validateBulk method that we expose for bulk creation, but it will take some work to refactor this, so I've just left it as an inline comment for later.

### Changes outside of the scope of this PR:
- Toast styling (will consolidate the styling of the toast when the design changes for the BulkIssue flow are ready)
- Notifications refactor to expose both `POST /notifications/bulks` and `POST /notifications/sample`
- `validateSingle` method for letters
- `singleCreate` transaction method can be deduplicated with the `bulkCreate` method

## Before & After Screenshots

**AFTER**:
Screen Recording 2023-07-27 at 10 05 52 AM
RPReplay_Final1690260388

## ***IMPORTANT: Deployment notes***
Sample letters are only created on template creation. For existing templates, we will need to create sample letters manually and add them to templates. 

This is already done on staging, but when we deploy this to production, we must add sample letters manually


frontend/src/features/issue/hooks/templates.hooks.ts
```diff
@@ -79,3 +81,35 @@ export const useCreateBulkLetterMutation = ({
...
+    {
+      onSuccess: () => {
+        toast({
+          title: 'Sample SMS sent',
+          status: 'success',
+        })
+        onSuccess?.()
+      },
+      onError: (e: WretchError) => {
+        console.log(e)
```
PikkaPikkachu, line 106: nit : we can remove this?

gweiying: [fbb1885](https://github.com/opengovsg/letters/pull/204/commits/fbb1885c97e6ea421baf56537178618a2b6a07cd)

backend/src/letters/letters.module.ts
```diff
@@ -19,7 +19,7 @@ import { LettersValidationService } from './letters-validation.service'
     TypeOrmModule.forFeature([User]),
     AuthModule,
     BatchesModule,
-    TemplatesModule,
+    forwardRef(() => TemplatesModule),
```
PikkaPikkachu, line 22: Oh no — we still have to add this to resolve the issue (just curious)?

backend/src/letters/letters.service.ts
```diff
@@ -79,20 +63,61 @@ export class LettersService {
...
+          entityManager,
+        )
+        const letterDto = {
+          ...securedLetter,
+          batchId: batch.id,
+          userId,
+          templateId: template.id,
+          shortLink: '',
+          isPasswordProtected: false,
+        }
```
PikkaPikkachu, line 106: Isn't this logic the same as `L147-L70ish` ? We can refactor, but non-blocking :))

PikkaPikkachu:
Overall LGTM ! There might be some scope for refactoring?

backend/src/letters/letters.service.ts
gweiying: Yep! There's some scope for refactoring the singleCreate function that's right

backend/src/letters/letters.module.ts
gweiying: Yep, mentioned briefly in the PR notes above. As TemplatesService uses LettersService to create sample letters and LettersController uses TemplatesService to find the relevant template, there's a module-level dependency between Templates and Letters that we couldn't avoid, but this should not cause any runtime issues!

backend/src/letters/letters.controller.ts
```diff
@@ -28,24 +28,56 @@ import {
...
-    const letters = await this.lettersService.bulkCreate(user.id, bulkRequest)
+    const { templateId } = bulkRequest
+    const template = await this.templatesService.findOne(templateId)
+    if (!template) throw new NotFoundException('template not found')
+    const letters = await this.lettersService.bulkCreate(
+      user.id,
+      template,
+      bulkRequest,
+    )
+    if (bulkRequest.phoneNumbers)
```
PikkaPikkachu, line 75: teeny nit (and also question): we can just extract the `phoneNumbers` above on `L67` from `bulkrequest` right? Will avoid the chaining here and on `L79`

PikkaPikkachu (approved):
LGTM, thanks!

backend/src/letters/letters.module.ts
AikoPath, line 22: We could leave a comment here explaining **why**  we are doing this here. Maybe just a link to the notion doc.

backend/src/notifications/sms-notifications.service.ts
```diff
@@ -17,6 +17,10 @@ export class SmsNotificationsService {
     private readonly config: ConfigService,
   ) {}
 
+  isValidPhoneNumber(recipient: string): boolean {
+    return !!recipient.match(ACCEPTED_SINGAPORE_PHONE_NUMBERS_REGEX)
+  }
+
```
AikoPath, line 23: I like the extraction of this regex matcher.
We could reuse this in the `validatePhoneNumbers` function in an upcomming refactoring.

backend/src/templates/templates.service.ts
```diff
@@ -1,32 +1,48 @@
...
     const template = this.repository.create(parsedTemplate)
-    return await this.repository.save(template)
+    const createdTemplate = await this.repository.save(template)
+
+    const sampleLetter =
+      await this.lettersService.generateSampleLetterForTemplate(
+        userId,
+        template,
+      )
+    const updatedTemplate = {
```
AikoPath, line 41: `templateWithSampleLetter`?

shared/src/constants/regex.ts
```diff
@@ -3,6 +3,8 @@ export const ACCEPTED_TEMPLATE_FIELDS_REGEX = /{{( *[a-zA-Z0-9_]+ *)}}/g
 export const ACCEPTED_SINGAPORE_PHONE_NUMBERS_REGEX =
   /^(\+65|65|65-|\+65-)?([89]\d{7})$/
 
+export const ACCEPTED_SINGAPORE_PHONE_NUMBERS_REGEX_WITHOUT_65 = /^([89]\d{7})$/
+
```
AikoPath, line 7: if there is a clean way of reusing this pattern we could include this in the pattern above.
(admittedly I can see how that might be significantly less readable)

AikoPath (approved):
Only some tiny refactoring ideas, that might or might not be addressed at a future time.

Thanks a lot for picking this up and fighting through the dependencies issue!

backend/src/templates/templates.service.ts
gweiying: abf1239fb842561312646a7a5af72acc75ea6916

backend/src/letters/letters.module.ts
gweiying: [11e52db](https://github.com/opengovsg/letters/pull/204/commits/11e52db00402788ecdd6fc2226d72c43783b94b9)

backend/src/letters/letters.controller.ts
gweiying: [fc06d6f](https://github.com/opengovsg/letters/pull/204/commits/fc06d6fb12dd0efbc3fdb3d324307d6adba4f545)

shared/src/constants/regex.ts
gweiying: Left it as is for now as I can't think of a significantly more elegant solution!

PR #205, opened 2023-07-27
chore: added frontend password check

YangXingHao96:
## Context

Currently there are work flow issues whenever there are emoji, Chinese characters in the password, after investigation, it seems like HTTP header password field of type iso-8859-1 (emoji, Chinese character) will trigger type error, this type error is not handled in our code. In order to avoid this, we added a validation logic at front end to perform regex to ensure that we handle this form of password exception

Closes [insert issue #]

## Approach

Regex using zod library
**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:
[insert screenshot here]

**AFTER**:
[insert screenshot here]

## Tests

## Deploy Notes

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:

## Risks


frontend/src/features/public/components/PasswordProtectedView.tsx
```diff
@@ -51,16 +53,34 @@ const BaseGridLayout = (props: GridProps) => {
   )
 }
 
+const schema = z.object({
+  password: z
+    .string()
+    .trim()
+    .regex(ACCEPTED_ASCII_PASSWORD_REGEX, { message: 'Invalid Password' }),
```
PikkaPikkachu, line 60: Just wanted to check that these are all validation, and not actually modifying the password ? 
Because of the user enters `ABC  ` and the password is `ABC` then they would be able to see the letter, but the correct behaviour should be to throw an error ?

PikkaPikkachu: (I am not super familiar with zod — so not sure if `trim` here is a validation check or actually trimming the string)

YangXingHao96: you are right, it actually does remove spaces, let me edit it accordingly

gweiying:
Chatted offline with @YangXingHao96 and suggested to remove the Controller + render logic, as we are able to use Input as a simple uncontrolled component. Will help simplify the maintenance of the component!

frontend/src/features/public/hooks/letters.hooks.ts
```diff
@@ -26,12 +26,13 @@ export const useGetLetterByPublicId = ({
...
-    () =>
-      api
+    () => {
+      return api
         .url(`/public/letters/${letterPublicId}`)
         .headers({ password })
         .get()
-        .json<GetLetterPublicDto>(),
+        .json<GetLetterPublicDto>()
+    },
```
gweiying, line 35: btw, non-blocking and just fyi! 

`() => a+b` is equivalent to `() => { return a+b }` in Javascript, as long as the code is a one-line function! So there's no difference between the two functions, but for readability if you prefer the latter that's also fine!

frontend/src/features/public/hooks/letters.hooks.ts
```diff
@@ -26,12 +26,13 @@ export const useGetLetterByPublicId = ({
     refetch: refetchLetter,
   } = useQuery<GetLetterPublicDto, PasswordResponseError>(
     publicQueryKeys.letters(letterPublicId),
```
gweiying, line 28: Hey @YangXingHao96 we might need to add `password` here as a query key dependency. The query is not refetching when the password is changed. The reason is `setPassword` is not running synchronously, so when we call refetch after `setPassword` in the flow below, the password is not guaranteed to be reset yet. This has the side effect of running a refetch before the password is set on the `useState`, see video below.

```
  const handleSubmit = async (password: string) => {
    setPassword(password)
    await refetchLetter()
  }

```
See the video below 
https://github.com/opengovsg/letters/assets/39231249/eac29da2-691a-4716-9032-c527bb4fabd5

gweiying (approved):
lgtm!

PR #208, opened 2023-07-27
feat: rating

disKeith:
## Context

This PR introduces a feature to collect feedback from users by adds a feedback form that allows end users to share their feedback on their experience of using the product

Closes [Notion Task](https://www.notion.so/opengov/Add-a-rating-for-issuing-flow-at-the-end-of-the-flow-and-nps-d6a41fb26d2c438ca82ce2cfbd95788f)

## Approach

1. Implement backend endpoint `{host}/api/feedback` that logs all incoming request bodies
  1a. Setup datadog metric and visualisations to visualise the logged feedbacks

2. Implement frontend feedback modal that calls the endpoint on submission
  2a. Trigger feedback modal at the end of bulk issue flow (when user clicks download links)
  2b. Add feedback button at navigation bar to allow users to send feedbacks outside of the issue flow

**Features**:

- Shared
  - Added `FeedbackDTO` class to be used as object for sending and logging feedbacks

- Backend
  - Added feedback module, controller, and service that receives POST requests for feedbacks
  - Logs `msg` as the message content, `feedback_rating` as the star rating, and `feedback_text` as the optional remark in the feedback submitted

- Frontend
  - Added `feedback` hook to handle calling the feedback endpoint and show toasts
  - Added a new component `SendFeedbackModal` that is used at `BulkIssueCompletionCard` and `UserMenu` components
    - Use react-hook-form to add controllers to store and parse:
      - `feedbackRating`: Star rating field that requires user to select a rating from 1 to 5 stars
        - validates if no rating was given
      - `feedbackText`: Optional text input for users to write any remarks
        - 500 character limit with character count
        - regex only accepts safe character sets (ref: [AWS Object key safe characters](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html))
  - Added `Feedback.svg` illustration for the `SendFeedbackModal` component
  - Updated `BulkIssueCompletionCard` to also trigger the `SendFeedbackModal` when download link is clicked
  - Updated `UserMenu` to have a new `Send Feedback` button to allow users to send feedback from the navbar

## Before & After Screenshots

**BEFORE**:

https://github.com/opengovsg/letters/assets/119388168/b4b6f44d-265a-41d8-87fb-dcf6bfa040b1

**AFTER**:

https://github.com/opengovsg/letters/assets/119388168/4aac9bed-1a71-4d3b-baf1-0aaebff81656


backend/src/feedback/feedback.service.ts
```diff
@@ -0,0 +1,21 @@
...
+  constructor(
+    @InjectPinoLogger(FeedbackService.name)
+    private readonly logger: PinoLogger,
+  ) {}
+
+  create(feedbackDto: FeedbackDto) {
+    this.logger.info({
+      msg: `Feedback (Rating: ${feedbackDto.rating}): ${feedbackDto.feedbackText}`,
+      feedback_rating: feedbackDto.rating,
+      feedback_text: feedbackDto.feedbackText,
```
PikkaPikkachu, line 17: This is more of a product decision comment/question -- but are we not interested in storing the officer's email who gave this feedback or any other information or is this more of an anonymous feedback ?

frontend/src/components/Navbar/components/UserMenu.tsx
```diff
@@ -1,42 +1,55 @@
...
 
 import { NavMenuItem } from './NavMenuItem'
 
 export const UserMenu = () => {
   const { adminUser, logout } = useAdminAuth()
+  const {
+    isOpen: isOpenFeedback,
+    onOpen: onOpenFeedback,
+    onClose: onCloseFeedback,
+  } = useDisclosure()
```
PikkaPikkachu, line 28: We don't necessarily have to define the names for 'isOpen', etc here, we did this in the other component because there were two modals (no change required -- but just sharing haha)

frontend/src/features/feedback/components/SendFeedbackModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+import { routes } from '~constants/routes'
+
+import { useSendFeedbackMutation } from '../hooks/feedback.hooks'
+
+interface SendFeedbackModalProps {
+  isOpen: boolean
+  onClose: () => void
+}
+
+interface FormData {
```
PikkaPikkachu, line 29: Nit: we could rename it to something more specific like FeedbackFormData or UserFeedbackFormData?

frontend/src/features/feedback/hooks/feedback.hooks.ts
```diff
@@ -0,0 +1,35 @@
...
+        onSuccess?.()
+        toast({
+          title: 'Thanks for your feedback!',
+          status: 'success',
+        })
+      },
+      onError: () => {
+        toast({
+          title: 'Error',
+          description: 'Failed to send feedback.',
```
PikkaPikkachu, line 29: Question -- Is this the title or description ? For above this is the title (for success)

PikkaPikkachu (commented):
Overall LGTM -- just some minor things :)

backend/src/feedback/feedback.service.ts
```diff
@@ -0,0 +1,21 @@
...
+import { FeedbackDto } from '~shared/dtos/feedback.dto'
+
+@Injectable()
+export class FeedbackService {
+  constructor(
+    @InjectPinoLogger(FeedbackService.name)
+    private readonly logger: PinoLogger,
+  ) {}
+
+  create(feedbackDto: FeedbackDto) {
```
PikkaPikkachu, line 13: minor nit: we could always avoid the chaining by extracting fields from `feedbackDto` like so:
```
const {rating, feedbackText ...} = feedbackDto
```

frontend/src/features/feedback/components/SendFeedbackModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+    defaultValues: {
+      feedbackRating: 0,
+      feedbackText: '',
+    },
+  })
+
+  const { mutateAsync, isLoading } = useSendFeedbackMutation({
+    onSuccess: () => {
+      onClose()
+      navigate(`/${routes.admin.index}/${routes.admin.templates.index}`)
```
gweiying, line 55: Suggestion, not sure if we want to redirect to home everytime somebody submits a feedback form. We might want to chain it to the `onClose` function in BulkIssueCompletionCard.tsx instead, for example 
```
<SendFeedbackModal
  isOpen={isOpen} 
  onClose={() => {
    onClose() 
    navigate(`/${routes.admin.index}/${routes.admin.templates.index}`)
  }
/>
`

File: frontend/src/features/feedback/components/SendFeedbackModal.tsx [hunk 7]
```diff
@@ -0,0 +1,168 @@
...
+                      cursor={'pointer'}
+                      onClick={() => field.onChange(rating)}
+                      boxSize={10}
+                      mr={1}
+                      as={field.value >= rating ? BsFillStarFill : BsStar}
+                    />
+                  ))}
+                </Flex>
+              )}
+            />
```
gweiying, line 113: nice component!

frontend/src/features/feedback/components/SendFeedbackModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+                        message: 'Exceeded character limit.',
+                      },
+                    })}
+                  />
+                  <FormHelperText textAlign={'right'}>
+                    {field.value.length} / 500
+                  </FormHelperText>
+                </>
+              )}
+            />
```
gweiying, line 150: Tiny suggestion, we don't actually need to use the Controller component here, since we can register `Textarea` neatly as a simple uncontrolled component!

```
          <FormControl isInvalid={!!errors.feedbackText} {...formStyles}>
            <Textarea
              isInvalid={!!errors.feedbackText}
              fontSize={'md'}
              placeholder={
                '(Optional) Tell us more about your experience, share your feature wish list, or provide suggestions on how to improve.'
              }
              resize={'none'}
              minHeight={'144px'}
              {...register('feedbackText', {
                pattern: {
                  value: /^[a-zA-Z0-9!-_.*'()\n ]+$/,
                  message:
                    'Only alphanumeric characters and certain special characters are allowed.',
                },
                maxLength: {
                  value: 500,
                  message: 'Exceeded character limit.',
                },
              })}
            />
            <FormHelperText textAlign={'right'}>
              {watch('feedbackText').length} / 500
            </FormHelperText>
            <FormErrorMessage>
              {errors.feedbackText && errors.feedbackText.message}
            </FormErrorMessage>
          </FormControl>
```

gweiying (commented):
Good stuff, thanks for looking into this task! The implementation was clean and well-scoped, easy to review 👍 

Apart from the questions from Prakriti, just one question from me. I might be missing something, but where is `feedback_name` on DD being specified?

backend/src/feedback/feedback.service.ts
gweiying, line 17: I might be missing something, but where is `feedback_name` on DD being specified?

disKeith: agreed that it might be useful to log the user which provides us a means to reach out to a specific user - it seems to me that an anonymous feedback approach seems okay for now as there's no explicit requirement / use for the user's details for ops (we can KIV this?)

frontend/src/components/Navbar/components/UserMenu.tsx
disKeith: ah, i defined the names there as the `isOpen` variable was already used when unpacked in Line 32 `{({ isOpen }) => (`

frontend/src/features/feedback/components/SendFeedbackModal.tsx
disKeith: 🫡 updated in https://github.com/opengovsg/letters/pull/208/commits/93da014c1dceb31f7ae8cf4131c3f8fa7d39615b

frontend/src/features/feedback/hooks/feedback.hooks.ts
disKeith: Nice catch! for parity, have removed description and moved it to `title` instead - updated in https://github.com/opengovsg/letters/pull/208/commits/93da014c1dceb31f7ae8cf4131c3f8fa7d39615b

backend/src/feedback/feedback.service.ts
disKeith: yeah that's a better practice - updated in https://github.com/opengovsg/letters/pull/208/commits/93da014c1dceb31f7ae8cf4131c3f8fa7d39615b

frontend/src/features/feedback/components/SendFeedbackModal.tsx
disKeith: that makes more sense! thanks, updated in https://github.com/opengovsg/letters/pull/208/commits/93da014c1dceb31f7ae8cf4131c3f8fa7d39615b

frontend/src/features/feedback/components/SendFeedbackModal.tsx
disKeith: right, thanks for catching this - it's indeed redundant, updated in https://github.com/opengovsg/letters/pull/208/commits/93da014c1dceb31f7ae8cf4131c3f8fa7d39615b

disKeith:
@PikkaPikkachu 
> Overall LGTM -- just some minor things :)
> 
> EDIT: Another generic comment, we actually don't give any option to officers to not submit the feedback and go back to homepage. Some officers might not be willing to give feedback, so we should still have a cancel button on the `feedbackModal` which redirects them back to home page ? (Might be more of product/design discussion though :) )

yeah agreed, not having any visual elements to allow me to close it feels a little uncomfortable to me (added it in https://github.com/opengovsg/letters/pull/208/commits/6874d5f0f7551f801f42e316202f313d411ae3ca and we can see what the team thinks)

@gweiying 
> Good stuff, thanks for looking into this task! The implementation was clean and well-scoped, easy to review 👍
> 
> Apart from the questions from Prakriti, just one question from me. I might be missing something, but where is `feedback_name` on DD being specified?

Ah i have updated DD metrics & visualisation to query based on `feedback_rating` instead - the reason for it being `feedback_name` previously was because we were still in discussion of possibly adding categories for the type of feedback, hence its declaration on DD

gweiying (commented):
lgtm!

gweiying (approved):
lgtm!

PR #209, opened 2023-07-28
chore: disable button while sending sample sms

gweiying:
## Context

We are able to resubmit the test sms form multiple times as the button is not disabled while the SMS is being sent

## Approach

Read the `isLoading` state from the `useMutation` and disable the form/button while loading.

## Tests

Send test SMS Button should not be clickable while sending SMS


AikoPath (approved):
Good catch!

PR #210, opened 2023-07-30
Feat: Send letter using email

PikkaPikkachu:
## Context

We are extending the existing letters service by allowing officers to notify citizens by sending emails. This PR closes this notion [task](https://www.notion.so/opengov/2ccadbea5c734aa09614b4ca8cc7177e?v=76cff9fca57b496f857c990e2639af2f&p=268965876d8f4cefa6fa43e4dbb650f0&pm=s). 

Refer to technical doc [here](https://www.notion.so/opengov/24072023-Email-Integration-with-LetterSG-81c31d1fccdb4a97b78cd5915b64f88d)

## Approach

Refer to technical doc [here](https://www.notion.so/opengov/24072023-Email-Integration-with-LetterSG-81c31d1fccdb4a97b78cd5915b64f88d)

**Features**:

- We are extending the existing functionality by allowing officers to send emails for notifying citizens about the issued letters.

## Before & After Screenshots

**AFTER**:

https://github.com/opengovsg/letters/assets/25703976/97922734-4708-4f66-8de3-5c6484de555a

**Regression Test for SMS Flow**

https://github.com/opengovsg/letters/assets/25703976/d1430b5e-1007-4b47-96e4-14e756a6ed9f

**Error Flow for Emails**
There are two types of errors possible for emails:
1. Invalid email
2. Missing email 

Screenshot 2023-07-31 at 2 55 56 PM

## Tests

1. Have added backend testing for the email integration feature
4. Manual testing for checking any regressions on the existing SMS feature.


frontend/src/utils/stringUtils.ts
```diff
@@ -7,3 +9,15 @@ export const pluraliseIfNeededForCount = (
...
   singular: string,
 ): string => (count > 1 ? `${singular}s` : singular)
+
+export const pluraliseCitizenNotificationMethod = (
+  citizenNotificationMethod: string,
+): string => {
+  return citizenNotificationMethod === CitizenNotificationMethod.SMS
+    ? 'SMSes'
+    : 'emails'
+}
```
AikoPath, line 19: This might return a wrong string in case a string other than `SMS` or `email` is passed in.

A solution could be to have the Enum as the parameter instead of any string.

I think a more flexible solution would even be to just extend `pluraliseIfNeeded` to support the correct plural of `SMS`.

shared/src/dtos/letters.dto.ts
```diff
@@ -36,6 +36,14 @@ export class CreateSingleLetterDto {
   phoneNumber?: string
 }
 
+export class CreateNotificationDto {
+  @IsDefined()
+  @IsArray()
+  recipients: string[]
+  @IsDefined()
+  notificationType: CitizenNotificationMethod
```
AikoPath, line 44: It would be good if we can use one terminology through out the code base to describe individual "things".
E.g. either `notification**Type**` or `..notification**Method**` or `notification**Channel**` etc. and use that everywhere unless there is a specific reason to use a different term.

AikoPath (commented):
Thanks!
Overall this looks good and as discussed the general approach makes a lot of sense, especially given the plan of separating the letter creation from the notification creation.

I just left some small comments regarding some naming etc.

frontend/src/features/issue/components/modals/SendSampleSmsModal.tsx
```diff
@@ -67,45 +78,65 @@ export const SendSampleSmsModal = ({
...
+              </Flex>
+            ) : (
+              <Flex width="70%">
+                <InputGroup borderColor="#BFC2C8">
+                  <Input
+                    type="email"
+                    placeholder="example@open.gov.sg"
+                    {...register('recipient', {
+                      pattern: {
+                        value: ACCEPTED_PUBLIC_OFFICER_EMAIL_REGEX,
```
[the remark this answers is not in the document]
PikkaPikkachu: Thanks, changed it!

backend/src/letters/letters.controller.ts
```diff
@@ -46,16 +46,18 @@ export class LettersController {
     @CurrentUser() user: User,
     @Body() issueSampleLetterDto: IssueSampleLetterDto,
   ) {
-    const { templateId, phoneNumber } = issueSampleLetterDto
+    const { templateId, notificationObject } = issueSampleLetterDto
```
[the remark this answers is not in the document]
gweiying: What do you guys think of `createNotificationDto`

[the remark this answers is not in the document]
PikkaPikkachu: Yup makes sense and is consistent with the rest of code base as well!

frontend/src/features/issue/components/BulkIssueSendLettersCard.tsx
```diff
@@ -96,12 +104,17 @@ export const BulkIssueSendLettersCard = ({
...
+          Send {capitalizeFirstLetter(citizenNotificationMethod ?? '')}
+        </Text>
       </HStack>
       <Text fontSize={'12px'}>SENDER NAME</Text>
       <Text textStyle={'body-1'}>LetterSG</Text>
       <Spacer padding={'12px'} />
-      <Text fontSize={'12px'}>SMS MESSAGE </Text>
+      <Text fontSize={'12px'}>
+        {citizenNotificationMethod?.toUpperCase()} MESSAGE{' '}
+      </Text>
```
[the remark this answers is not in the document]
PikkaPikkachu: I think we will be doing through some copy changes for email body as well! Probably something we can address in later PR — when we do final QA for this ? What do you think ?

backend/src/notifications/email-notifications.service.ts
```diff
@@ -0,0 +1,56 @@
...
+    return true
+  }
+
+  async sendEmailNotification(
+    templatedMessages: { message: string; letterId: number }[],
+    recipients: string[],
+  ): Promise<Partial<Notification>[]> {
+    const sentEmails: Partial<Notification>[] = []
+
+    await Promise.all(
```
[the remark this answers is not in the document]
PikkaPikkachu: Ahh makes sense, we might also wanna change for SMS flow ? That also uses `Promise.all` :(

backend/src/notifications/email-notifications.service.ts
```diff
@@ -0,0 +1,56 @@
...
+      templatedMessages.map(async ({ message, letterId }, index) => {
+        const recipient = recipients[index]
+        try {
+          const mail = {
+            to: recipient,
+            from: `${this.config.get('otp.sender_name')} <${this.config.get(
+              'otp.email',
```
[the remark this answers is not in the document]
PikkaPikkachu: This is actually not the logged in user. We are getting the credentials for letters email sending domain, which is the same as the credentials through which OTP email is sent. I think we should rename the config variables, so that it's not confusing why we are fetching opt related info here for sending emails.

backend/src/notifications/email-notifications.service.ts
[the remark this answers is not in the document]
gweiying: Didn't review the rest of the PR but just a small comment from me here on this, I actually swapped back from `Promise.allSettled()` to `Promise.all()` on the SMS flow on purpose because I thought that `Promise.allSettled()` enforced a very opinionated way of returning the results of the array of Promises which we shouldn't have to deliberately work around for.

So instead of running `Promise.allSettled()` and iterating through the settled results to log for all failed promises, I just ran with `Promise.all` and put a try-catch in to catch and log the errors. If @AikoPath doesn't feel strongly about it I think using `Promise.all` might be simpler!

backend/src/notifications/email-notifications.service.ts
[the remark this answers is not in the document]
AikoPath: Oh, I see. Yes in that case we should definitely rename that as I was already confused :)

backend/src/notifications/email-notifications.service.ts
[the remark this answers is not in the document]
AikoPath: No concerns from my side. That also works 👍

frontend/src/features/issue/components/BulkIssueSendLettersCard.tsx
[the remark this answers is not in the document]
AikoPath: Yea, that works. As long as we don't forget about it.

shared/src/dtos/letters.dto.ts
PikkaPikkachu: Ohh I think that would require some renaming to SMS code flow as well. We can pick this up and the copy changes as task for one or two PRs on top of this ?

frontend/src/features/issue/components/modals/SendSampleSmsModal.tsx
```diff
@@ -67,45 +76,66 @@ export const SendSampleSmsModal = ({
...
+                    {...register('recipient', {
+                      validate: {
+                        validOfficerEmail: (email) =>
+                          isGovSgOrWhitelistedEmail(email) === true ||
+                          'Email should be a valid public officer email',
+                      },
+                    })}
+                  />
+                </InputGroup>
+              </Flex>
```
gweiying, line 124: nice! reusing this validation logic here makes sense :D

backend/src/notifications/email-notifications.service.ts
```diff
@@ -0,0 +1,56 @@
...
+  constructor(
+    private readonly mailerService: MailerService,
+    @InjectPinoLogger(EmailNotificationService.name)
+    private readonly logger: PinoLogger,
+    private readonly config: ConfigService,
+  ) {}
+
+  isInEmailAllowlist(recipient: string): boolean {
+    return true
+  }
```
gweiying, line 20: After using this in staging for the last few weeks, I've realized how invaluable this feature is, thanks @AikoPath for the original suggestion. Let's implement this for email mode too!

backend/src/notifications/notifications.service.ts
```diff
@@ -20,15 +27,28 @@ export class NotificationsService {
   async sendNotifications(
     userId: number,
     letters: Letter[],
-    recipients: string[],
+    createNotificationDto: CreateNotificationDto,
```
gweiying, line 30: Just flagging this for consideration, I don't think we need to change anything here for now, but wondering about the naming here as this object is meant for both single and bulk notifications, but it isn't entirely intuitive to me from the name that this can be used for both!

backend/src/notifications/notifications.service.ts
```diff
@@ -20,15 +27,28 @@ export class NotificationsService {
...
+      case CitizenNotificationMethod.EMAIL: {
+        const sentEmails =
+          await this.emailNotificationService.sendEmailNotification(
+            templatedMessages,
+            createNotificationDto.recipients,
+          )
+        return this.notificationRepository.save(sentEmails)
+      }
+      default:
+        return []
```
gweiying, line 50: just for consideration, should we be throwing an error here instead? not relevant for now since the code is relatively lean and simple, but might make for easier debugging in the future instead of failing silently.

gweiying (approved):
nice stuff, looks very clean!

what do you think about implementing the email whitelist before the feature is released? so far based on our tests on staging, it feels much safer with the whitelist!

frontend/src/features/issue/components/BulkIssueUploadCsvCard.tsx
```diff
@@ -163,6 +179,9 @@ export const BulkIssueUploadCsvCard = ({
     if (isSendViaSms && phoneNumbers.length === 0) {
       addMissingParamErrorMessage('Phone Number')
     }
+    if (isSendViaEmail && emails.length === 0) {
```
gweiying, line 182: nit, missing check:
```
    if (!isSendViaEmail && emails.length > 0) {
      addExtraParamErrorMessage('Email')
    }
```

backend/src/notifications/notifications.service.ts
PikkaPikkachu: Hmm makes sense, makes sense!

backend/src/notifications/notifications.service.ts
PikkaPikkachu: Hmm, yup makes sense, so maybe something like `createNotificationsDto`? I think we can have another PR for renaming + copy changes in the code base!

PikkaPikkachu:
@gweiying thanks! 
> what do you think about implementing the email whitelist before the feature is released? so far based on our tests on staging, it feels much safer with the whitelist!

Implemented!

backend/src/config/config.schema.ts
```diff
@@ -287,4 +288,10 @@ export const schema: Schema<ConfigSchema> = {
     format: String,
     default: '',
   },
+  emailAllowList: {
+    doc: 'Allow list for email testing. Use "" to block all numbers, "*" to allow all numbers, or comma-separated emails, like so "abc@open.gov.sg, def@gov.sg"',
```
gweiying, line 292: tiny nit, typos here! "Use "" to block all ~numbers~ emails, "*" to allow all ~numbers~ emails"

gweiying (approved):
just two small typos otherwise everything looks ggggg

PR #212, opened 2023-07-31
Chore/ignore read from admin view

YangXingHao96:
## Context

Existing admin user can access the various letters from the admin dashboard, however every access will make the status of the letter to be read. This is not a wanted behaviour, in order to handle that, all letter links that come out from the admin page has additional param ?admin=true. When the admin field is set to true, the letter read state will not be updated.

For more information pls refer to the notion document:
https://www.notion.so/opengov/20230728-Ignore-read-from-admin-view-6d7559dd98c74ee4a063b0daece9d43a

Closes [insert issue #]

## Approach

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:
[insert screenshot here]

**AFTER**:
[insert screenshot here]

## Tests

## Deploy Notes

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:

## Risks


backend/src/public/public.controller.ts
```diff
@@ -13,19 +13,18 @@ import { GetLetterPublicDto } from '~shared/dtos/letters.dto'
...
     @Param('publicId') publicId: string,
   ): Promise<GetLetterPublicDto> {
     const letter = await this.lettersService.findOneByPublicId(
       publicId,
       password,
     )
     if (!letter) throw new NotFoundException('Letter not found')
-
-    await this.lettersService.recordFirstReadAtById(letter.id)
+    await this.lettersService.recordFirstReadAtById(letter.id, adminAcess)
```
PikkaPikkachu, line 27: nit: typo `adminAccess` ?

backend/src/letters/letters.service.ts
```diff
@@ -225,19 +225,28 @@ export class LettersService {
...
-    const letter = await this.repository.findOneBy({ id })
-    if (!letter) throw new NotFoundException('Letter not found')
+  async recordFirstReadAtById(id: number, adminAccess?: boolean) {
+    // the admin value received is of type string from the http request's header
+    let admin = false
+    if (adminAccess) {
+      if (typeof adminAccess === 'string') {
+        admin = adminAccess === 'true'
+      }
+    }
```
PikkaPikkachu, line 235: hmm quick question, if `adminAccess` is of the type `string`, then why do we pass `adminAccess` as boolean in the function params? 
Also if we know if it's of the type params, couldn't we just simply do: 
```
const admin = adminAccess === 'true'
```
?

PikkaPikkachu: Also for some logic simplification we can just do : 
```
if(!adminAccess === 'true'){
    return 
}
// rest of the logic for recording first read
```

frontend/src/features/public/LetterPublicPage.tsx
```diff
@@ -20,6 +20,15 @@ export const LetterPublicPage = (): JSX.Element => {
   const [passwordInstructions, setPasswordInstructions] = useState('')
   const transformScale = useTransformScale()
   const [letterHeight, setLetterHeight] = useState<number | undefined>(0)
+  const queryString = window.location.search
+  const urlParams = new URLSearchParams(queryString)
+  let admin = false
+  if (urlParams.has('admin')) {
```
PikkaPikkachu, line 26: nit: do we need to check for `.has`, I guess `.get` would just return `undefined` if it does not have any `admin` param , which would be handled by the second `if` for string check?

PikkaPikkachu (commented):
Overall LGTM :)
Just a few nits and comments on string conversion and handling booleans.

gweiying:
Just dropping a short message here, realized we can use the session cookie received on the backend request to check if the user is logged in, if logged in we can ignore the read request. Then we wouldn't have to do anything on the frontend to change the url parameterization! What do you think is better? Let's chat tmr morning to discuss this.

gweiying:
>Just dropping a short message here, realized we can use the session cookie received on the backend request to check if the user is logged in via the @CurrentUser() decorator. If logged in we can ignore the read recording. This would mean that we wouldn't have to do anything on the frontend to change the url parameterization.

Discussed offline and decided to use the session cookie instead, as its simpler and with the dashboard revamp the officers are more likely to retrieve the letter link from the downloaded csv than from the dashboard (until the individual batch view is implemented)

frontend/src/features/public/hooks/letters.hooks.ts
```diff
@@ -29,7 +29,9 @@ export const useGetLetterByPublicId = ({
     () => {
       return api
         .url(`/public/letters/${letterPublicId}`)
-        .headers({ password })
+        .headers({
+          password,
+        })
```
gweiying, line 34: tiny nit, think this is leftover from some linting?

package.json
```diff
@@ -52,4 +52,4 @@
     "class-transformer": "^0.5.1",
     "class-validator": "^0.14.0"
   }
-}
+}
```
gweiying, line 55: tiny nit, think this is leftover from some linting?

frontend/src/features/dashboard/IssuedLettersPage.tsx
```diff
@@ -17,7 +17,6 @@ import { GetLetterDto } from '~shared/dtos/letters.dto'
 import { getLetterPublicLink } from '~utils/linkUtils'
 
 import { IssuedLetterDrawer } from './components/IssuedLetterDrawer'
```
gweiying, line 19: tiny nit, think this is leftover from some linting?

backend/src/public/public.controller.ts
```diff
@@ -24,8 +27,7 @@ export class PublicController {
       password,
     )
     if (!letter) throw new NotFoundException('Letter not found')
-
-    await this.lettersService.recordFirstReadAtById(letter.id)
+    await this.lettersService.recordFirstReadAtById(letter.id, user)
```
gweiying, line 30: suggestion, what do you think about moving the `if (!user)` check here instead, so that the `recordFirstReadAtById` logic is kept very simple?

backend/src/core/decorators/current-user.decorator.ts
```diff
@@ -9,3 +9,13 @@ export const CurrentUser = createParamDecorator(
...
+
+export const OptionalUser = createParamDecorator(
+  (_data: unknown, ctx: ExecutionContext) => {
+    const { session } = ctx.switchToHttp().getRequest<Request>()
+    if (!session.user) {
+      return undefined
+    }
+    return session.user
+  },
+)
```
gweiying, line 21: looks good!

backend/src/public/public.controller.ts
YangXingHao96: Edited this part of the code, it seems that the rest is indeed due to some linting issue. I have run npm run lint:fix before pushing though

backend/src/public/public.controller.ts
```diff
@@ -6,16 +6,19 @@ import {
   Param,
 } from '@nestjs/common'
 import { mapLetterToPublicDto } from 'core/dto-mappers/letter.dto-mapper'
+import { User } from 'database/entities/user.entity'
```
gweiying, line 9: tiny nit, this import should be changed to `'../database/entities/user.entity'`

PR #214, opened 2023-08-01
fix: upload csv validation

disKeith:
## Context

This PR fixes the following validation scenario during bulk issue CSV uploads (for headers):

When user uploads a csv with valid headers, but missing row values in 1 or more rows, the column validation shows `{{header}} is missing` - which is not intended and should be shown as an error message in row validation

Closes [Notion Task](https://www.notion.so/opengov/bug-Throwing-missing-column-instead-of-row-level-validation-fd333680c5ad447480a047bc2f3d3485)

## Approach
`BulkIssueUploadCsvCard` 
- Extracted column validation logic to `useParseCsv` hook
    - Added logic to append `Password` or `Phone Number` fields to `templateFields` if the option is selected - used when downloading samples and field validation in `useParseCsv`
    - Removed `getErrorMessageArr` function and replaced with only handling errors from `uploadCsv` & `revalidateCsv` instead
    - Added useEffect hook to revalidate an existing uploaded csv if issue settings changed

`csvUtils`

- Updated `csvToJsonArr` method to return `result.meta.fields` to be able to extract the parsed headers even when a csv has only 1 row with missing values, which previously didn't allow the possibility of extracting all header values in that scenario

`useParseCsv`
- Updated `useParseCsv` to require `templateFields` parameter for column validation
- Added `validateFields` function to handle validation during uploads and revalidation
- Updated `parseCsv` function to `uploadCsv`
  -  Extract parsed csv fields even when there are no corresponding rows through `resolvedFields`
  - Updated `Your csv file is empty` error to show when csv is empty, and when the csv has valid fields but no rows
  -  Validate the resolved fields
- Added `revalidateCsv` function which revalidates the existing csv's state


gweiying:
Discussed offline on the PR, pasting the notes:
> - GOING IN A GOOD DIRECTION! It's nice that all the validation logic is removed from BulkIssueUploadCsvCard so that now all the validation happens inside useParseCsv. The component logic is much cleaner and the validations are consolidated.
> - we should expose a revalidate function from the hook
>   - why: if the value of isPasswordProtected, isSendViaSms changes, we need to retrigger the validation, so that the column level validation will rerun
>   - considerations: we probably don't wanna reparse the whole csv, so we might want to think about storing additional values inside the useParseCsv hook e.g. maybe storing the output of the parsed file, and just rerun the validation on the stored state
> - other small bugs:
>   - noticed that the downloaded sample had two phone number columns
>   - probably related to the validation above not rerunning, but I was able to submit a csv successfully with "manually send links" option selected, but which contained phone numbers!

gweiying (approved):
god damn! nice refactor! 100/100

PR #219, opened 2023-08-03
feature: landing page UI revamp

PikkaPikkachu:
## Context

Revamp Dashboard UI.
Allow users to download CSVs for letters issued in the same batch.

Closes [this issue](https://www.notion.so/opengov/2ccadbea5c734aa09614b4ca8cc7177e?v=76cff9fca57b496f857c990e2639af2f&p=9dcc5ebe811d41a190a3d1588bb3f246&pm=s).

## Approach

New endpoint in `batches` controller 
* GET/ - fetches list of batches for authenticated user
```
{
    batches: [{
        batchId: number;
        templateName: string;
        createdAt: string;
        totalCount: number;
        readCount: number;
    }];
    count: number;
}
```
* GET/:batchId - fetches all letters in a batch
```
{
    templateName: string;
    createdAt: string;
    letters: [{
          publicId: string;
          fieldValues?: string;
          recipient?: string;
          firstReadAt: string;
      ]};
}
```

* New UI

https://github.com/opengovsg/letters/assets/25703976/ab3c547a-31b1-447c-8018-7f2fdebbfa4e


backend/src/batches/batches.service.ts
```diff
@@ -33,12 +40,78 @@ export class BatchesService {
...
+      },
+    })
+
+    if (!batches) throw new NotFoundException('Batches not found for this user')
+    return batches
   }
 
-  findOne(id: number) {
-    return `This action returns a #${id} batch`
+  async findAll(
```
gweiying, line 62: `nit`, this function name isn't very indicative of the function since we're technically not finding all batches - it's a paginated retrieval. what do you think of `findAndCount` or something along those lines

backend/src/batches/batches.service.ts
```diff
@@ -33,12 +40,78 @@ export class BatchesService {
...
+    userId: number,
+    limit: number,
+    offset: number,
+  ): Promise<GetBatchDto> {
+    const [batches, count] = await this.findAndCountAllByUserId(
+      userId,
+      limit,
+      offset,
+    )
+    const batchesData = await Promise.all(
```
gweiying, line 72: nit, I might be missing it but it doesn't seem like a Promise is necessary here since we're not doing anything asynchronous?

backend/src/batches/batches.service.ts
```diff
@@ -33,12 +40,78 @@ export class BatchesService {
     return batch
   }
 
-  findAll() {
-    return `This action returns all batches`
+  async findAndCountAllByUserId(
```
gweiying, line 43: suggestion, entirely non-blocking! I'm wondering what do you think of collapsing this function into the logic below? There's not much that this function does in addition to retrieving batch data for the user id, and I'm not sure where else this function would be reused.

backend/src/batches/batches.service.ts
```diff
@@ -33,12 +40,78 @@ export class BatchesService {
...
+          item.firstReadAt != null ? sum++ : 0
+          return sum
+        }, 0)
+        return {
+          batchId: batch.id,
+          templateName: batch.templateName,
+          createdAt: batch.createdAt.toDateString(),
+          totalCount: letters ? letters.length : 0,
+          readCount: readCount,
+        } as GetBatchesResponseDto
```
gweiying, line 85: just thinking over this return type, since we already have letters we could also return `letters` here and standardise the return type for the `findWithPagination`, so that you can have just a single DTO here, e.g. `GetBatchDto` and `GetBatchDto[]`

gweiying: That would also mean that on the frontend, we could use the existing `letters` object for download unless it doesn't exist, and use the `findOne` endpoint otherwise. Would prevent another refetch on the backend, since we're already fetching all batches above.

gweiying, line 85: Can consider implementing a `mapBatchToDto` function here similar to `mapLetterToDto`

gweiying: Another side comment especially if we might want to do the single DTO type here `GetBatchDto`, we could consider implementing `mapBatchToDto` similar to the `mapLetterToDto`. Would help reuse the logic.

backend/src/notifications/notifications.service.ts
```diff
@@ -24,6 +24,16 @@ export class NotificationsService {
...
+  async getRecipient(letterId: number): Promise<string | null> {
+    const notification = await this.notificationRepository.findOneBy({
+      letterId: letterId,
+    })
+    if (!notification) {
+      return null
+    }
+    return notification.recipient
+  }
+
```
gweiying, line 36: nit, this looks like an unused function?

shared/src/dtos/batches.dto.ts
```diff
@@ -3,6 +3,33 @@ export class UpdateBatchDto {}
...
+export class GetBatchDto {
+  batches: GetBatchesResponseDto[]
+  count: number
+}
+
+export class GetBatchesByIdResponseDto {
+  templateName: string
+  createdAt: string
+  letters: BatchLetter[]
+}
```
gweiying, line 35: Left a comment above in `batches.service.ts` about simplifying the DTOs! I find the naming of these DTOs unintuitive, for example `GetBatchDto` sounds like a DTO for a single batch, but returns multiple batches, whereas `GetBatchesByIdResponseDto` sounds like a DTO for multiple batches, but returns a single batch. Hopefully with simplifying the DTOs above we might be able to resolve this confusion

gweiying (commented):
Overall the PR looks great! Left some inline nits mostly about naming and code cleanup ideas. 

Another thing, I would love to get your help with a simple load test. Since we're using lazy loading to load all letters in the batches for the `findWithPagination` endpoint, would be good to verify that the load time and server load when we fetch 8 batches x (500 letters + 500 notifications) is reasonable. I don't think it'd be a big problem but let's just test it out to be safe!

backend/src/batches/batches.controller.ts
```diff
@@ -22,13 +30,17 @@ export class BatchesController {
...
+    @Query('offset', ParseIntPipe) offset: number,
+  ) {
+    return this.batchesService.findAll(user.id, limit, offset)
   }
 
   @Get(':id')
-  findOne(@Param('id') id: string) {
-    return this.batchesService.findOne(+id)
+  async findOne(@Param('id') id: string): Promise<GetBatchesByIdResponseDto> {
+    return await this.batchesService.findOne(+id)
```
gweiying, line 43: mostly just FYI, it's a weird Javascript thing but actually the `async` and `await` are not necessary here. 

The function is returning a resolved promise directly without awaiting any asynchronous operation within the function body. The `findOne` method of the batchesService is expected to be an asynchronous function itself (since it returns a Promise<GetBatchesByIdResponseDto>), but the function does not need to await it because it is simply returning the promise directly.

shared/src/dtos/batches.dto.ts
```diff
@@ -1,8 +1,26 @@
...
   passwordInstructions: string
 }
+export class GetBatchesResponseDto {
+  batchId: number
+  templateName: string
+  createdAt: string
+  totalCount: number
+  readCount: number
+  letters: BatchLetter[]
+}
```
gweiying, line 16: tiny nit, can we swap the name for this to `GetBatchDto` ? This is a single object that's returned, so seems like the names are currently swapped.

shared/src/dtos/batches.dto.ts
```diff
@@ -1,8 +1,26 @@
...
+export class BatchLetter {
+  publicId: string
+  fieldValues?: string
+  recipient?: string
+  firstReadAt: string
+}
+export class GetBatchDto {
+  batches: GetBatchesResponseDto[]
+  count: number
+}
```
gweiying, line 26: can we swap the name for this to `GetBatchesResponseDto`? This is returning multiple batches so makes sense to swap this to `GetBatchesResponseDto`!

gweiying (approved):
lgtm! just one final tiny nit on naming

shared/src/dtos/batches.dto.ts
PikkaPikkachu: Thanks! Makes sense, revised here [b9f585e](https://github.com/opengovsg/letters/pull/219/commits/b9f585e76025a12ebecdc995b1fc3901d28962de)

gweiying (approved):
lgtm!

gweiying (approved):
thanks for fixing the batch name so quickly!

PR #221, opened 2023-08-03
fix: routing for non excistent template id

ddanelynn:
## Context

Currently, when accessing a non-existent template id route, the frontend just stays on the issue letter dashboard page instead of throwing an error. 
Accessing the frontend route for issuing a non-existent template id, e.g. http://letters-stg.beta.gov.sg/admin/templates/100000/issue should redirect a user back to Templates dashboard.

Closes [this issue](https://www.notion.so/opengov/2ccadbea5c734aa09614b4ca8cc7177e?v=76cff9fca57b496f857c990e2639af2f&p=400b4197f04f44c9954e7bf801cf226b&pm=s).

## Approach
Add a `TemplateIdProtectedRoute` layer under `AdminProtectedRoute` to ensure template Id in url is valid.

**Issue**
There will be a 4-5s delay before page is redirected to /templates due to backend fetching speed (checking if the templateId exists).

* Non-existent template Id

https://github.com/opengovsg/letters/assets/66818697/67a1e5c4-8a9c-4445-b7a2-8c6ad6d927b1

* Existent template Id

https://github.com/opengovsg/letters/assets/66818697/c1e4f9ce-6bbc-4372-8caf-536e8ae9c6f2


frontend/src/features/TemplateIdProtectedRoute.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+export const TemplateIdProtectedRoute = ({
+  children,
+}: {
+  children?: JSX.Element
+}) => {
+  const navigate = useNavigate()
+  const location = useLocation()
+  const [isTemplateIdValid, setIsTemplateIdValid] = useState<boolean>(false)
+  const templateId = +location.pathname.split('/')[3]
```
PikkaPikkachu, line 15: nit: did we try using this for fetching the template id? 
```
const { templateId } = useTemplateId()
```

frontend/src/features/TemplateIdProtectedRoute.tsx
```diff
@@ -0,0 +1,39 @@
...
+      return
+    }
+    if (!isTemplatesLoading && !template) {
+      navigate('templates')
+    }
+    setIsTemplateIdValid(true)
+  }, [template, navigate, isTemplatesLoading])
+
+  return (
+    <>
```
PikkaPikkachu, line 29: nit: I don't think we should need `<></>` here?

PikkaPikkachu (commented):
Overall LGTM, just a few suggestions :))

gweiying:
Took a quick look at the PR and happy to discuss further on this!

I'm not sure we would need a wrapping component `TemplateIdProtectedRoute` here instead of directly checking for the template's existence in `BulkIssuePage`. Off the top of my head, I'm actually not sure where else we would use this wrapper, so I would think that the simpler approach is preferable. What do you think?

frontend/src/features/issue/BulkIssuePage.tsx
```diff
@@ -31,50 +33,78 @@ export const BulkIssuePage = (): JSX.Element => {
...
+            </VStack>
+          </VStack>
+        </FormProvider>
+      ) : (
+        <Center height="100vh">
+          <Spinner size="xl" />
+        </Center>
+      )}
+    </>
+    // eslint-disable-next-line @typescript-eslint/unbound-method
```
PikkaPikkachu, line 108: nit: what's this es-lint comment ignoring 😅 ?

PikkaPikkachu:
Honestly, I wouldn't mind having the template validation check to be a separate component (like we had before), since `BulkIssuePage` component is pretty big in itself, so we could actually have a separate parent component, but not have it at the router level, instead at `BulkIssuePage` level — overall I think this is okay too though :)

LGTM! Thanks for the quick iterations! :D

frontend/src/features/issue/BulkIssuePage.tsx
ddanelynn: oops

frontend/src/features/TemplateIdProtectedRoute.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+  useEffect(() => {
+    if (isTemplatesLoading) {
+      return
+    }
+    if (!isTemplatesLoading && !template) {
+      navigate('templates')
+    }
+    setIsTemplateIdValid(true)
+  }, [template, navigate, isTemplatesLoading])
```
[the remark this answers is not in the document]
ddanelynn: I think there is no error returns from the hook, not super sure how to use it :000

frontend/src/features/TemplateIdProtectedRoute.tsx
ddanelynn: Tried to delete but it gave me red lines. I think need a wrapper for conditional returns

frontend/src/features/TemplateIdProtectedRoute.tsx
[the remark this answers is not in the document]
PikkaPikkachu: const { template } = useGetTemplateById(templateId)
if (error) {
  return <Navigate to={`/templates`} />
}
try this?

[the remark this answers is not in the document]
PikkaPikkachu: Ahh there's no error being returned by the hook.

frontend/src/features/issue/BulkIssuePage.tsx
```diff
@@ -31,50 +33,77 @@ export const BulkIssuePage = (): JSX.Element => {
...
+
+  useEffect(() => {
+    if (isTemplatesLoading) {
+      return
+    }
+    if (!isTemplatesLoading && !template) {
+      navigate('templates')
+    }
+    setIsTemplateIdValid(true)
+  }, [template, navigate, isTemplatesLoading])
```
PikkaPikkachu, line 49: do we need to have `navigate` as a dependency in the `useEffect`?

PikkaPikkachu (approved):
LGTM! Can we also check that the templates with existing or valid IDs are routed to correct, the test plan does not have a screen recording for that — so just wanted to make sure, we tested both the cases :)

frontend/src/features/TemplateIdProtectedRoute.tsx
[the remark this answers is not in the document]
gweiying: Hey all, what do you think about modifying the `useGetTemplateById` hook here in `templates.hooks.ts` which would prevent us from relying on this `useEffect` workaround instead? This is a common design pattern for React Query hooks, so we should lean into it and use the exposed error instead of using the `useEffect` which can cause some unnecesary rerenders on the page. 

We also do the same logic in `useGetLetterByPublicId`

```
export const useGetTemplateById = (templateId: number) => {
  const { data, isLoading, error } = useQuery(['templates', templateId], () =>
    api.url(`/templates/${templateId}`).get().json<GetTemplateDto>(),
  )

  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
  return { template: data!, isTemplateLoading: isLoading, error }
}
```

gweiying (changes requested):
Hey Dannelyn! I think you're on the right track with this loading spinner, if you're interested you can also check our react suspense to see how to do this elegantly, but I don't think it's necessary for now otherwise. 

I'm not so sure about using this useEffect and useState to handle the `isTemplateValid` logic, because `useEffect` often causes alot of unnecessary rerenders on the frontend that can cause the application to be non-performant. Since we can just use the returned errors from the `react-query` component to conditionally redirect, we should just use that error instead of engineering a new workaround. Let's chat on Monday if it's unclear so far!

frontend/src/features/issue/BulkIssuePage.tsx
```diff
@@ -31,8 +33,26 @@ export const BulkIssuePage = (): JSX.Element => {
...
+      </Center>
+    )
+  }
+  if (error) {
+    console.log('Template Id error')
+  }
+
+  if (!isTemplatesLoading && !template) {
+    navigate('templates')
+  }
```
gweiying, line 53: hey @ddanelynn, just wondering why we chose to handle the error and the navigation separately? are there different cases that `error` and `!isTemplatesLoading && !template` are returning?

ddanelynn: ah make sense

frontend/src/features/issue/hooks/templates.hooks.ts
```diff
@@ -20,12 +20,12 @@ export const useTemplateId = (): { templateId: number } => {
 }
 
 export const useGetTemplateById = (templateId: number) => {
-  const { data, isLoading } = useQuery(['templates', templateId], () =>
+  const { data, isLoading, error } = useQuery(['templates', templateId], () =>
```
gweiying, line 23: Just tested it out locally, we might also want to add a `retry: false` here ([see docs](https://tanstack.com/query/v4/docs/react/reference/useQuery)). The spinner is loading for a long time, and that's probably because the query is retrying a few times before failing.

gweiying (approved):
lgtm! behaves beautifully, thanks for the change

PR #222, opened 2023-08-04
Feat/dashboard UI revamp typeorm

ddanelynn:
## Context

Change API endpoints from letters to batches.


backend/src/batches/batches.service.ts
```diff
@@ -51,12 +59,50 @@ export class BatchesService {
...
+        } as GetBatchesResponseDto
+      }),
+    )
+    return { batches: batchesData, count: count }
   }
 
-  findOne(id: number) {
-    return `This action returns a #${id} batch`
+  async findOne(id: number): Promise<GetBatchesByIdResponseDto> {
+    const batch = await this.repository.findOneBy({ id: id })
```
PikkaPikkachu, line 92: teeny nit: would just  {id} work ?

PikkaPikkachu:
LGTM! Let's merge it into main branch and do a set of E2E on localhost and then push to staging!

PikkaPikkachu (approved):

backend/src/batches/batches.controller.ts
```diff
@@ -4,9 +4,13 @@ import {
   Delete,
   Get,
   Param,
+  ParseIntPipe,
   Patch,
   Post,
+  Query,
 } from '@nestjs/common'
+import { CurrentUser } from 'core/decorators/current-user.decorator'
```
PikkaPikkachu, line 12: I think some tests are failing because of the incorrect import here — it should be :

```
 import { CurrentUser } from '../core/decorators/current-user.decorator'
 ```

File: backend/src/batches/batches.controller.ts [hunk 3]
```diff
@@ -4,9 +4,13 @@ import {
...
+import { User } from 'database/entities'
```
PikkaPikkachu, line 13: And this one should be : `import { User } from '../database/entities'`

PR #224, opened 2023-08-07
fix: email sending fix for logging + throttling

PikkaPikkachu:
## Context

During load testing we observed that emails can't be send and are rate limited. Post further discussion with Alexis, we have decided to throttle these email sending. 
The other fix that this PR addresses is to save the email notification to DB even when the email is not allowlisted for consistency.

## Approach

1. Install `promise-throttle`. Followed the implementation in this [PR for highway](https://github.com/opengovsg/highway/pull/225/commits/4481a5d1a392ac0f61975f4a0c4dd1441dc764bf).
2. Add logging to DB even when email is not allowlisted.

## Before & After Screenshots
**AFTER**:

https://github.com/opengovsg/letters/assets/25703976/5c1d4470-af51-4936-80ed-964e0abeaa97

## Tests

1. Did load testing locally, was able to recieve multiple emails to docker email. 
2. Sent not allowlisted email, they were still saved to backend DB.


backend/src/config/config.schema.ts
```diff
@@ -292,6 +292,6 @@ export const schema: Schema<ConfigSchema> = {
     doc: 'Allow list for email testing. Use "" to block all emails, "*" to allow all emails, or comma-separated emails, like so "abc@open.gov.sg, def@gov.sg"',
     env: 'EMAIL_ALLOW_LIST',
     format: String,
-    default: '',
+    default: 'prakriti@open.gov.sg',
```
gweiying, line 295: tiny nit, might want to remove this haha!

backend/src/notifications/email-notifications.service.ts
```diff
@@ -36,6 +36,13 @@ export class EmailNotificationService {
         const isEmailAllowListed = this.isInEmailAllowlist(recipient)
         if (!isEmailAllowListed) {
           this.logger.info(`Sending mail to ${recipient}`)
+          // even when not sending emails save to DB
+          sentEmails.push({
+            channel: NotificationChannel.EMAIL as string,
+            message,
+            letterId,
+            recipient,
+          })
```
gweiying, line 45: small suggestion, we might want to move this outside of the if-else check instead of duplicating the `Array.push` method in lines 61-65

gweiying (approved):
lgtm! tiny nits but otherwise looks fine

backend/src/notifications/email-notifications.service.ts
PikkaPikkachu: Thanks fixed here: [9bf7ecf](https://github.com/opengovsg/letters/pull/224/commits/9bf7ecf3c03de696e7fe417212d7519242074d2d)

backend/src/config/config.schema.ts
PikkaPikkachu: Fixed here, [9bf7ecf](https://github.com/opengovsg/letters/pull/224/commits/9bf7ecf3c03de696e7fe417212d7519242074d2d)!

PR #226, opened 2023-08-08
fix: add error message when csv file exceeds 500 rows

ameliatay:
## Context

Currently, on the 'Upload CSV' page when users wish to notify recipients (route admin/templates/1/issue), there is a requirement that CSV files must have a maximum of 500 rows. However error message is not shown to users when CSV files with more than 500 rows are uploaded

Closes [this issue](https://www.notion.so/opengov/bug-not-correctly-throwing-error-when-more-than-500-letters-are-being-created-at-once-df63726c160c42d5bf6257ea4df4e5da?pvs=4)

## Approach

Add additional check when users upload files to check `resolvedData` length where 1 row equates to 1 element in `resolvedData`'s array

## Before & After Screenshots

**BEFORE**:
https://github.com/opengovsg/letters/assets/83273830/59903839-145c-4e9d-bd2b-dfbe3b645d16

**AFTER**:
https://github.com/opengovsg/letters/assets/83273830/0c0ecd8c-90cd-4306-ba63-f8417959bfa5


PikkaPikkachu:
Thank you for putting up the fix!

While this works for 500 rows error case specifically, we still are missing out on resolving the issue more generically for other errors that are being silently thrown in the backend a.k.a :
```
if (letterParamMaps.length > BULK_MAX_ROW_LENGTH) {
      return {
        success: false,
        message: 'Number of rows exceeded max length of bulk create',
      }
    }

    if (passwords && letterParamMaps.length !== passwords?.length) {
      return {
        success: false,
        message: 'Number of passwords does not match number of letters',
      }
    }

    if (
      passwordInstructions &&
      (passwordInstructions.length < MIN_PASSWORD_INSTRUCTION_LENGTH ||
        passwordInstructions.length > MAX_PASSWORD_INSTRUCTION_LENGTH)
    ) {
      return {
        success: false,
        message:
          'Password instructions need to have a minimum of 10 characters and maximum number of 560 characters',
      }
    }

    if (recipients && letterParamMaps.length !== recipients?.length) {
      return {
        success: false,
        message: 'Number of recipients does not match number of letters',
      }
    }
```
i.e. that do not return an `errors` array. I am open to picking that change as a separate PR — however, we could ideally tackle them in one shot :))

What do you all think ?

frontend/src/features/issue/hooks/useParseCsv.tsx
```diff
@@ -82,6 +82,10 @@ const useParseCsv = (templateFields: string[]) => {
       return setError(['Your csv file is empty'])
     }
 
+    if (resolvedData.length > 500) {
```
gweiying, line 85: Tiny nit, might want to import `BULK_MAX_ROW_LENGTH` from `shared/src/constants/letters.ts` to avoid hardcoding this

ameliatay:
> Thank you for putting up the fix!
> 
> While this works for 500 rows error case specifically, we still are missing out on resolving the issue more generically for other errors that are being silently thrown in the backend a.k.a :
> 
> ```
> if (letterParamMaps.length > BULK_MAX_ROW_LENGTH) {
>       return {
>         success: false,
>         message: 'Number of rows exceeded max length of bulk create',
>       }
>     }
> 
>     if (passwords && letterParamMaps.length !== passwords?.length) {
>       return {
>         success: false,
>         message: 'Number of passwords does not match number of letters',
>       }
>     }
> 
>     if (
>       passwordInstructions &&
>       (passwordInstructions.length < MIN_PASSWORD_INSTRUCTION_LENGTH ||
>         passwordInstructions.length > MAX_PASSWORD_INSTRUCTION_LENGTH)
>     ) {
>       return {
>         success: false,
>         message:
>           'Password instructions need to have a minimum of 10 characters and maximum number of 560 characters',
>       }
>     }
> 
>     if (recipients && letterParamMaps.length !== recipients?.length) {
>       return {
>         success: false,
>         message: 'Number of recipients does not match number of letters',
>       }
>     }
> ```
> 
> i.e. that do not return an `errors` array. I am open to picking that change as a separate PR — however, we could ideally tackle them in one shot :))
> 
> What do you all think ?

Thanks Prakriti for pointing that out! Is it okay if I try the change and have you review it? Actually implemented that catch previously but thought it was a messier way to solve this specific issue so removed it. Hoping to see if what I did is what you had in mind! If it's different, then yes let'e open a separate ticket 👍🏻

gweiying:
Agree with what @PikkaPikkachu said, though I think tackling it in separate PRs could make sense!

PikkaPikkachu:
Yeah makes sense @ameliatay ! In my understanding there are two solutions:
1. We move all backend checks to frontend like you did for this one 
2. Or we improvise the existing solution for handling errors from backend [I have a teeny bit of preference for this one]

Let's not block this PR though and we can pick that up separately in parallel! Thanks!

gweiying:
Agree with what @PikkaPikkachu mentioned! I think it doesn't hurt to duplicate the frontend and backend validation for some of the simple checks for example max length < 500, min length !== 0, but I think overall would be good to improve our frontend/backend error handling. Can also combine that with the [cleanup of unused validations here](https://www.notion.so/opengov/chore-Unused-length-validation-checks-f42fdc991b7e41f0a5ea84fd96dd9738).

ameliatay:
Got it! Updated PR to remove hardcoded value, we can merge this first work on the rest separately then. Thank you both!

frontend/src/features/issue/hooks/useParseCsv.tsx
```diff
@@ -82,6 +83,10 @@ const useParseCsv = (templateFields: string[]) => {
       return setError(['Your csv file is empty'])
     }
 
+    if (resolvedData.length > BULK_MAX_ROW_LENGTH) {
+      return setError(['Maximum of 500 rows has been exceeded'])
```
gweiying, line 87: nit, might want to use `Maximum of ${BULK_MAX_ROW_LENGTH} rows has been exceeded` too to avoid hardcoding the numerical value in the error message too.

PikkaPikkachu (approved):
LGTM, thank you for the quick iterations! 
(Remember to squash and merge :-) )

PR #228, opened 2023-08-08
fix: email validation regex

ddanelynn:
## Context

During issue flow with email (with and without password protect) - uploading an invalid email passes the csv validation

CSV

```
keywords,curly,Password,Email
A,b,1a2b3c4d5f,testuser@open.gov.sg
C,d,1a2b3c4d5f,testuser@op
```

`testuser@op` should show `1. Email should be a valid email address`

Closes [this issue](https://www.notion.so/opengov/2ccadbea5c734aa09614b4ca8cc7177e?v=76cff9fca57b496f857c990e2639af2f&p=8a48bb1cb3964790b80decf336e51a54&pm=s)

https://github.com/opengovsg/letters/assets/66818697/289f0ab7-efe1-443b-89e1-c67ab8c653a7

## Approach
Fix email regex string.


PikkaPikkachu:
LGTM, I think we were previously declaring `.` as optional in the regex which is this part `?:\.`

Thanks for the quick fix!

PikkaPikkachu (approved):

PR #229, opened 2023-08-08
feat: add user guide to user menu and change user menu item style

ameliatay:
## Context

Users face difficulty to locating the letters guide as it's nested in the drop down. A more visible way to navigate letters is required for better experience

Closes [this issue](https://www.notion.so/opengov/Adding-user-guide-contact-us-to-nav-bar-db4d1fb8b073457787bed479d5b716bb?pvs=4)

## Approach

Added a new button to user menu that opens a new tab () when clicked. Changed the styling for both buttons 'Get Help' and 'Share Feedback'

## Before & After Screenshots

**BEFORE**:
Screenshot 2023-08-08 at 1 10 37 PM

**AFTER**:
https://github.com/opengovsg/letters/assets/83273830/685989e6-e1ba-469f-b716-b4b1ee30867c


PikkaPikkachu:
@gweiying are we planning this as part of this week's release? If so, then we should also push this change to staging, so we can get any feedback if required from the product team?

Change overall LGTM! :))

frontend/src/components/Navbar/components/UserMenu.tsx
```diff
@@ -31,9 +31,22 @@ export const UserMenu = () => {
...
+            >
+              Guide
+            </Button>
+            <Button
+              size="sm"
+              variant="ghost"
+              border="none"
+              onClick={onOpenFeedback}
+            >
+              Contact
```
gweiying, line 49: Tiny nit, let's change this to `Share feedback` since it links to the feedback modal :D It's not directly a contact us link. 

I think the Screenshots on the task were confusing, sorry about that.

frontend/src/components/Navbar/components/UserMenu.tsx
```diff
@@ -31,9 +31,22 @@ export const UserMenu = () => {
...
         <>
-          <HStack spacing={10}>
-            <Button size={'sm'} variant={'outline'} onClick={onOpenFeedback}>
-              Share Feedback
+          <HStack spacing={6}>
+            <Button
+              size="sm"
+              variant="ghost"
+              border="none"
+              onClick={() => window.open(USER_GUIDE, '_blank')}
```
gweiying, line 39: tiny nit here, what do you think of using the <Link> component from Chakra instead of the onClick `window.open` Javascript function?

two reasons why I'd prefer the Chakra component, one for standardization with the way we handle links in the frontend, and another because the Chakra link component is more likely to offer better accessibility compared to reusing a Button as a link, since Link is designed as a navigation component!

gweiying:
> If so, then we should also push this change to staging, so we can get any feedback if required from the product team?
Agree, good idea! Let's update the copy and the behavior and get it on staging too :)

PikkaPikkachu (approved):
LGTM! We will push the change to staging with the rest of final changes!

PR #231, opened 2023-08-09
chore: email body for sending emails

PikkaPikkachu:
## Context

This PR adds email body when the emails are sent by public officer. Initially when we integrated emails, we used a vanilla text, this PR adds some html formatting to the emails being sent with a change in the copy. 

## Approach

Since at this point, both emails and SMSes would have different templates, we decided to move the `getTemplateMessage` to the email and SMS service respectively and let each of these classes handle the logic for template generation separately. 

The PR fixes some failing tests as well due to the change in logic. 

## Before & After Screenshots

**BEFORE**:
Screenshot 2023-08-10 at 4 05 11 AM

**AFTER**:
Screenshot 2023-08-10 at 2 56 33 AM

Email Preview
Screenshot 2023-08-10 at 11 39 11 AM

SMS Preview
Screenshot 2023-08-10 at 11 39 36 AM


gweiying (commented):
changes look good! we might want to update the preview text here as well:

Screenshot 2023-08-10 at 10 19 59 AM

PikkaPikkachu:
> changes look good, just one small catch, we might want to update the preview text here as well:

Thanks missed this! Updated previews for both post discussion with Feli!

gweiying (approved):
changes look good to me, thanks for the quick iteration! 

side thought, though we can make this optimisation change outside the scope of this PR. we could move the sms and email message out into the shared folder so that both the backend notifications service and the frontend display can use it, which will help in the future to prevent any copies from getting out of sync

PikkaPikkachu:
> side thought, though we can make this optimisation change outside the scope of this PR. we could move the sms and email message out into the shared folder so that both the backend notifications service and the frontend display can use it, which will help in the future to prevent any copies from getting out of sync

Sounds good — we might have to take care of formatting it properly on frontend :))

PR #232, opened 2023-08-09
chore: log provider id for emails

PikkaPikkachu:
## Context

This PR adds support for recording the message ID of the email being sent in the providerID column in the notifications table. 
This is useful for debugging and tracking individual emails that were sent out if required. 

## Approach

1. For dev mode: we log a constant string + UUID, for e.g. `DEV_MODE_NO_EMAIL_SENT_37373803-bdca`
2. For prod mode: we get the message ID from the email response and log it. 
3. For dev mode, with email allow listed: `<05d8412a-dd8b-51f8-25df-f4638e652a76@mail.postman.gov.sg>` something similar to this gets recorded.

## Before & After Screenshots

**AFTER**:

DB records post sending emails: 
Screenshot 2023-08-10 at 4 08 52 AM

API call for tracking email: `https://api.postman.gov.sg/v1/transactional/email/4075904`


gweiying (commented):
hey @PikkaPikkachu, small suggestion here, happy to discuss whether this makes sense. 

The approach looks sound but I was wondering if we could standardize the behavior of email-sending with sms-sending. For SMS-sending, anything that's not sent (whether it's because there's no Twilio credentials or because it's not on the allowlist) gets the `DEV_MODE_NO_SMS_SENT` tag. In this case, I would think that it would make sense for any emails that aren't sent to have the same `DEV_MODE_NO_EMAIL_SENT` tag regardless of whether they're not on the allowlist or because mailer used is just the maildev mock. Not sure if there's value having a separate tagging system or whether it would just be more developer-overhead to maintain!

gweiying:
> The approach looks sound but I was wondering if we could standardize the behavior of email-sending with sms-sending. For SMS-sending, anything that's not sent (whether it's because there's no Twilio credentials or because it's not on the allowlist) gets the DEV_MODE_NO_SMS_SENT tag. In this case, I would think that it would make sense for any emails that aren't sent to have the same DEV_MODE_NO_EMAIL_SENT tag regardless of whether they're not on the allowlist or because mailer used is just the maildev mock. Not sure if there's value having a separate tagging system or whether it would just be more developer-overhead to maintain!

Synced on this offline and we've decided not to change the behavior for now. Since `nodemailer` can be also reused in the future if we bring in our own mailserver, it doesn't make sense to change the behavior of `nodemailer` in our `dev` environment. Since the third scenario above only happens in local development, we will also not see the `<05d8412a-dd8b-51f8-25df-f4638e652a76@mail.postman.gov.sg>` in our staging or prod DBs for now. We'll add inline comments for developers to help reduce the developer context overhead!

gweiying (approved):
lgtm!

PR #237, opened 2023-08-11
Feat: detailed dashboard view UI 

PikkaPikkachu:
## Context

Build a detailed dashboard view of the batch letters for the new UI revamp, as in the figma [here](https://www.figma.com/file/KVFsGATLUfvPqKE2lY9niH/LetterSG?node-id=1634%3A55391&mode=dev). 

Read doc here for more [context](https://www.notion.so/opengov/LetterSG-UI-Revamp-Technical-Doc-c8fc21f3a57f4b0b954dc4417d496f9b).

## Approach

1. Build new components for the detailed batch view page. 
2. We also remove the old components that not being used anymore. 
3. This is mostly a frontend change, with a bit of change in the dto objects for further compatibility and reducing code base's complexity. 

**Features**:

1. When users click on the batch, they are redirected to the detailed batch letters page. 
2. They are shows certain stats about the batch. 
3. Officers can see the individuals letters that they issued and click on any of the rows to open a drawer for more information on the letters. 

## Before & After Screenshots

**AFTER**:

https://github.com/opengovsg/letters/assets/25703976/047e1332-c815-485c-9f5c-46d3bfef67f3

## Potential Future Work:
1. Save pagination state
2. Filters / search functionality 
3. Pagination of letters


backend/src/core/dto-mappers/letter.dto-mapper.ts
```diff
@@ -24,6 +24,10 @@ export const mapLetterToDto = (letter: Letter): GetLetterDto => {
     firstReadAt: letter.firstReadAt?.toDateString(),
     isPasswordProtected: letter.isPasswordProtected,
     issuedLetter: letter.issuedLetter,
+    fieldValues: letter.fieldValues,
+    recipient: letter.notifications?.length
+      ? letter.notifications[0].recipient
```
gweiying, line 29: suggestion, would be good to leave a comment here why we're returning the recipient from the first notification as a workaround!

frontend/src/features/dashboard/hooks/dashboard.hooks.ts
```diff
@@ -14,6 +15,22 @@ export const useGetTemplates = () => {
...
 }
 
+export const useBatchId = (): number => {
+  const { batchId } = useParams()
+  if (!batchId) throw new Error('No batchId provided')
+  return Number(batchId)
+}
+
+export const useGetBatchData = (batchId: number) => {
+  const { data, isLoading } = useQuery(['batches'], () =>
```
gweiying, line 25: nit, we would probably want to use a different query key here, perhaps `['batch', batchId]`, to make full use of the caching logic. using `['batches']` as a query key here will override the cache in `useGetBatches`!

backend/src/batches/batches.service.ts
```diff
@@ -64,7 +67,10 @@ export class BatchesService {
 
   async findOne(id: number): Promise<GetBatchDto> {
     const batch = await this.repository.findOne({
-      relations: { letters: true },
+      relations: {
+        letters: { notifications: true, template: true },
+        template: true,
+      },
```
gweiying, line 73: See comment above about fetching `template` twice!

backend/src/database/migrations/1691859222806-update-is-password-protected-and-notification-channel-columns-in-batches.ts
```diff
@@ -0,0 +1,36 @@
...
+      SET "isPasswordProtected" = TRUE
+      WHERE "id" IN (SELECT "batchId" FROM letters WHERE "isPasswordProtected"=TRUE GROUP BY "batchId")`,
+    )
+    // updates the "notificationChannel" column in the batches table according to the type of notification sent for the first letter
+    await queryRunner.query(
+      `UPDATE BATCHES 
+      SET "notificationChannel" = (SELECT "channel" from (SELECT "batchId", "channel" FROM letters l LEFT JOIN notifications n ON l."id" = n."letterId"
+      ) s WHERE "batchId" = "id" LIMIT 1)`,
+    )
+  }
```
gweiying, line 22: the migration scripts themselves look ok to me, but just wanna think about performance of the backfill migration. would be good to practice on staging and have a plan for the following:
- backup 
- monitor performance of migration (RDS CPU utilization, IOPs etc)
- checking that columns are filled correctly after migration

If the performance is poor on staging we might need to consider running the next release at an offpeak time or batching the backfill

frontend/src/features/dashboard/components/DownloadCsvButton.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { jsonArrToCsv } from '~utils/csvUtils'
+import { getLetterPublicLink } from '~utils/linkUtils'
+
+export interface DownloadCsvButtonProps {
+  batch: GetBatchDto
+  hideIcon?: boolean
+  size?: string
+}
+
+export const DownloadCsvButton = ({
```
gweiying, line 15: nice component!

frontend/src/features/dashboard/hooks/dashboard.hooks.ts
```diff
@@ -14,6 +15,22 @@ export const useGetTemplates = () => {
   return { templates: data!, isTemplatesLoading: isLoading }
 }
 
+export const useBatchId = (): number => {
+  const { batchId } = useParams()
+  if (!batchId) throw new Error('No batchId provided')
+  return Number(batchId)
+}
+
+export const useGetBatchData = (batchId: number) => {
```
gweiying, line 24: tiny nit, non-blocking. what do you think about renaming this to `useGetBatchById` for consistency with other hooks?

frontend/src/features/dashboard/components/batchViewComponents/IssuedBatchView.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { ErrorPage } from '~features/public/ErrorPage'
+
+import { IssuedBatchViewDetails } from './IssuedBatchViewDetails'
+import { IssuedBatchViewHeader } from './IssuedBatchViewHeader'
+import { IssuedBatchViewLetterTable } from './IssuedBatchViewLetterTable'
+
+export const IssuedBatchView = (): JSX.Element => {
+  const batchId = useBatchId()
+  const { batchDetails: batch } = useGetBatchData(batchId)
+  if (!batch) {
```
gweiying, line 16: Might need to add a check here for whether the batch is still loading, e.g. `if (!batch && !isBatchLoading) { ... }` 

I had some weird behavior locally where it showed the `ErrorPage` for a second before the batch loaded. If it's loading, we might need a loading state instead of `IssuedBatchViewLetterTable`.

Also, wondering what the ideal behavior should be, should we redirect the user back to `/admin/letters`, redirect the user to `/error`, or just display the `<ErrorPage />`? I would prefer the first, redirecting them back automatically `letters` as they should only reach this page if the batch id doesn't exist, but I don't have a strong preference here. What do you think?

gweiying (changes requested):
Did one pass to review the files so far, and mostly looks good to me! Left some small nits on cleanup as well as query/fetch performance on both frontend and backend

frontend/src/features/dashboard/components/batchViewComponents/IssuedBatchView.tsx
PikkaPikkachu: I think either way is fine for me — I think I have come across both behaviours of websites :
1. "Oops, 404, page not found" or something?
2. Where there is a automatic redirection

Sometimes I think redirection can be confusing, because I would try to access the same URL again, in hopes that it would lead me to the correct page, but an error page, shows that this thing doesn't exist and I wouldn't try again. 

But again neutral to either ways [+ seems like a rare case for now]. 

So let's just go with redirection then ?

PikkaPikkachu: As discussed offline, we will stick to redirection, since that's what we are using across the website for admin access!

PikkaPikkachu:
Thank you @gweiying for the review, have addressed the suggestions you proposed and as discussed offline, the DB/CPU performance seemed fine during our first staging deployment.

gweiying (approved):
lgtm~ thanks for the quick fixes

frontend/src/features/dashboard/components/batchViewComponents/IssuedBatchView.tsx
```diff
@@ -0,0 +1,54 @@
...
+import {
+  useBatchId,
+  useGetBatchById,
+} from '~features/dashboard/hooks/dashboard.hooks'
+
+import { IssuedBatchViewDetails } from './IssuedBatchViewDetails'
+import { IssuedBatchViewHeader } from './IssuedBatchViewHeader'
+import { IssuedBatchViewLetterTable } from './IssuedBatchViewLetterTable'
+
+export const IssuedBatchView = (): JSX.Element => {
```
gweiying, line 15: up for discussion, what do you think of moving this file out to `~features/dashboard` since it's a page view (no longer a component)?

PikkaPikkachu: changed [here](https://github.com/opengovsg/letters/pull/237/commits/e229033d7cc65dc30d920fe7cb855257ab10f961), thanks!

PR #239, opened 2023-08-14
fix: invalidate batches query on bulk create

disKeith:
## Context

This PR addresses a potential lag when issuing a letter batch - then viewing it in the dashboard, where the newly-issued batch may take some time to appear even after successfully issuing the batch.

Closes [Notion Task](https://www.notion.so/opengov/2ccadbea5c734aa09614b4ca8cc7177e?v=76cff9fca57b496f857c990e2639af2f&p=a1684658439047ef919054dfb851132f&pm=s)

## Approach

Invalidate `batches` query after successful bulk create in `useCreateBulkLetterMutation` to refetch in the background


frontend/src/features/issue/hooks/templates.hooks.ts
```diff
@@ -76,6 +76,7 @@ export const useCreateBulkLetterMutation = ({
       onSuccess: async (res: GetBulkLetterDto[]) => {
         // invalidate letters dashboard queries
         await queryClient.invalidateQueries(['letters'])
```
gweiying, line 78: do we still need to invalidate letters here?

also wondering if it would be good into using object types for query and query caching like with `publicQueryKeys` e.g.:

```
export const publicQueryKeys = {
  base: 'public' as const,
  letters: (letterPublicId: string, password?: string) =>
    [publicQueryKeys.base, letterPublicId, password] as const,
}
```

this will help us to avoid accidental typos with strings

gweiying (approved):
lgtm! can tackle the query keys object in a separate PR

frontend/src/features/issue/hooks/templates.hooks.ts
disKeith: > Do we still need to invalidate letters here?

nope! not needed for the new batch view (left it in cus the old `IssuedLetterPage` page was still in the code base) - but saw that it'll be deleted in #237

removed in https://github.com/opengovsg/letters/pull/239/commits/eeedc1a5170e919fffb5e18c3446d1e12023a215

PR #240, opened 2023-08-14
feat: backend test for letters

ddanelynn:
## Context

Unit test for `LettersService`
* Tests for single creation of letter with and without password
* Tests for bulk creation of letter with and without password
* Test for failed validation bulk creation

Unit test for `LettersController`
* Tests for bulk creation of letters with notification and password
* Tests for bulk creation of letters without notification and password


PikkaPikkachu:
Overall LGTM! Since this is a draft, I am thinking you'd be adding more test cases for letters service — i.e. password protected letters vs. non-password protected letters :))

backend/src/letters/letters.service.spec.ts
```diff
@@ -0,0 +1,362 @@
...
+      jest.spyOn(dataSource, 'transaction').mockResolvedValue([
+        {
+          anchorText: mockAnchorText,
+          issuedLetter: mockIssuedLetter,
+          fieldValues: mockFieldValues,
+          batchId: mockBatchId,
+          userId: mockUser.id,
+          templateId: mockTemplate.id,
+          shortLink: '',
+          isPasswordProtected: true,
```
PikkaPikkachu, line 321: small nit : I think this object is the same as one on `L333` then we can just declare this as a separate const and use it, we wouldn't need to re-declare.

backend/src/letters/letters.service.spec.ts
```diff
@@ -0,0 +1,362 @@
...
+      expect(result).toEqual([
+        {
+          anchorText: mockAnchorText,
+          issuedLetter: mockIssuedLetter,
+          fieldValues: mockFieldValues,
+          batchId: mockBatchId,
+          userId: mockUser.id,
+          templateId: mockTemplate.id,
+          shortLink: '',
+          isPasswordProtected: false,
```
PikkaPikkachu, line 282: tiny nit: actually just realised [reviewing the PR bottom up], we can make this a global const if possible and use it across all test cases, instead of having to redeclare separately. We could call is something like `createdLetterEntity` or `mockLetterEntity` ?

PikkaPikkachu:
Overall, LGTM, leaving some small suggestions for clubbing same objects into one var :)
Thanks for adding the comprehensive test cases!

PikkaPikkachu (approved):

PR #241, opened 2023-08-15
Fix/img load

gweiying:
## Context

The current scaling logic works by getting the height of the letter, and resizing it for smaller screen resolutions. However, we were running into issues when the letter height was set prematurely before the images had loaded. When the letter images were loading, they took up 0 height. However, after they loaded, the actual letter overflowed the bounds of the set letter, causing issues w the display.

## Approach

Run the `setLetterHeight` callback after the images of the html letter had loaded, or after `maxTimeToWaitInMs` (to avoid issues with corrupted images), whichever is earlier. `maxTimeToWaitInMs` is currently set to 1s which behaves reasonably for now (the assumption is that all non-corrupt images loads within 1s), but might need to adjust in production later~

## Before & After Screenshots

**BEFORE**:
localhost_3000_c8bao-s8pwy-ropil-src6u (1) Medium

**AFTER**:
localhost_3000_c8bao-s8pwy-ropil-src6u Medium


ddanelynn (approved):
lgtm!

frontend/src/utils/htmlUtils.ts
```diff
@@ -33,3 +33,34 @@ export const convertHtmlToPdf = (
     })
   })
 }
+
+export const onHtmlImgsLoaded = (
```
PikkaPikkachu, line 37: nit: non-blocking should we actually just name it `onHtmlImagesLoaded` for consistency across variable names for all `images`?

frontend/src/utils/htmlUtils.ts
```diff
@@ -33,3 +33,34 @@ export const convertHtmlToPdf = (
...
+      loadedImages++
+      if (loadedImages === images.length) {
+        callback()
+      }
+    } else {
+      images[i].onload = () => {
+        loadedImages++
+        if (loadedImages === images.length) {
+          callback()
+        }
```
PikkaPikkachu, line 62: tiny nit: can't we just move this logic outside of the `if` and `else`, so that we don't have to check inside both conditions?

frontend/src/utils/htmlUtils.ts
PikkaPikkachu: We can probably add some comments for what this function does? Saw that all other functions in this file also have some comments :))

frontend/src/utils/htmlUtils.ts
```diff
@@ -33,3 +33,34 @@ export const convertHtmlToPdf = (
...
+  const images = node.getElementsByTagName('img')
+  if (images.length === 0) {
+    callback()
+    return
+  }
+
+  let loadedImages = 0
+  setTimeout(callback, maxTimeToWaitInMs) // run callback after maxTimeToWaitInMs to avoid issues with corrupt images never loading
+
+  for (let i = 0; i < images.length; i++) {
```
PikkaPikkachu, line 51: alternatively, we could avoid the use of `i` var and do something like: 
```
for (const image of images) { }
```
but again up to you :))

PikkaPikkachu:
Left some tiny non-blocking suggestions! Otherwise logic seems intact! :))

frontend/src/utils/htmlUtils.ts
disKeith, line 62: relevant to the above comment, wdyt of something like this?

```suggestion
 for (const image of images) {
  image.complete ? loadedImages++ : (image.onload = () => loadedImages++);
  loadedImages === images.length && callback();
}
```

frontend/src/utils/htmlUtils.ts
gweiying: I thought we would standardise with the way images are referenced inside html, as in `` but don't feel strongly about this! Will rename then

frontend/src/utils/htmlUtils.ts
gweiying: Looks clean, but I think the `if (loadedImages === images.length) { callback() }` check needs to happen as part of the `onload` callback for each image so that the check will be triggered, since this isn't an asynchronous function

gweiying: Hmm, I believe the `if (loadedImages === images.length) { callback() }` check needs to happen as part of the `onload` callback for each image so that the check will be triggered, since this isn't an asynchronous function

gweiying: Can think further on how to optimise this~

gweiying: Can think further on how to optimise this~

frontend/src/utils/htmlUtils.ts
gweiying: [b337b5d](https://github.com/opengovsg/letters/pull/241/commits/b337b5df46c2d8f6efb2edc7600e358eb1d8af9c)

gweiying:
Refactored the callback mechanism here: [c61fa77](https://github.com/opengovsg/letters/pull/241/commits/c61fa77af8235142caa6b2d17e927857ae808dec) @disKeith @PikkaPikkachu let me know what you think~

PikkaPikkachu (approved):
LGTM! :))

PR #244, opened 2023-08-17
Chore/clean bulk issue validation

ameliatay:
## Context

1. Ensure all FE checks are also present independently on BE for external APIs in the future
2. Remove redundant checks (duplicated checks) and clean up error messages
3. Ensure any errors uncaught in the FE and passed from BE's validation can be shown to users for their action

Closes [this issue](https://www.notion.so/opengov/chore-Clean-up-bulk-issue-validation-errors-f42fdc991b7e41f0a5ea84fd96dd9738?pvs=4)

## Approach

1. Double check all FE checks are done on BE - checks such empty CSV had to be ported over
2. Remove duplicated checks - checks that are never reached as other checks cover their cases
3. Adjust how FE parses the response from BE letters/validate API to accommodate for 3 types of errors - general CSV errors, column-level errors, row-level errors

Above required changing the dto structure

**Improvements**:

- Previously error messages would be returned in 2 possible places in the API response (`errors` & `message`) so we will be collapsing it into a uniform errors array

**Bug Fixes**:

- Non row-level CSV errors caught in the BE were not populating on the FE UI

## Tests

Unit tests to be updated*


backend/src/letters/letters.service.ts
```diff
@@ -52,13 +52,16 @@ export class LettersService {
   ): [] {
     const {
       letterParamMaps,
+      letterColumnHeaders,
       passwords,
       passwordInstructions,
       createNotificationDto,
     } = createBulkLetterDto
+    console.log('entered validate')
```
PikkaPikkachu, line 60: nit: can be removed?

backend/src/letters/letters.service.ts
```diff
@@ -126,13 +129,16 @@ export class LettersService {
   ): Promise<Letter[]> {
     const {
       letterParamMaps,
+      letterColumnHeaders,
       passwords,
       passwordInstructions,
       createNotificationDto,
     } = createBulkLetterDto
+    console.log('entered create')
```
PikkaPikkachu, line 137: nit: same as above

frontend/src/features/issue/components/BulkIssueSendLettersCard.tsx
```diff
@@ -62,12 +65,37 @@ export const BulkIssueSendLettersCard = ({
...
+  const onError = (errors: BulkLetterValidationError[]) => {
+    setUploadCsvRowErrors(
+      errors
+        .filter(
+          (error) =>
+            error.errorType === BulkLetterValidationErrorType.ROW_ERROR,
+        )
+        .map((error) => error as BulkLetterValidationRowLevelError),
+    )
+    setUploadCsvNonRowErrors(
```
PikkaPikkachu, line 85: nit: might be preferrable to call it what these errors refer to for .e.g just `setUploadCsvErrors` instead of what it's not (that's just my thinking though)

frontend/src/features/issue/components/BulkIssueUploadCsvCard.tsx
```diff
@@ -111,18 +117,44 @@ export const BulkIssueUploadCsvCard = ({
...
+    setUploadCsvNonRowErrors(
+      errors
+        .filter(
+          (error) =>
+            error.errorType !== BulkLetterValidationErrorType.ROW_ERROR,
+        )
+        .map((error) => {
+          if (error.errorType === BulkLetterValidationErrorType.COLUMN_ERROR)
+            return `${
+              (error as BulkLetterValidationColumnLevelError).columnName
```
PikkaPikkachu, line 153: This logic seems to be the same as the one in `frontend/src/features/issue/components/BulkIssueSendLettersCard.tsx`, would be good to refactor it out, so we are not maintaining two error handling logics ?

frontend/src/features/issue/hooks/useParseCsv.tsx
```diff
@@ -26,29 +26,29 @@ const useParseCsv = (templateFields: string[]) => {
...
+    // // Check for fields that exist in the uploaded csv, but not in the template
+    // csvFields
+    //   .filter(
+    //     (key) => key !== '__parsed_extra' && !templateFields.includes(key),
+    //   )
+    //   .forEach((extraField) =>
+    //     errorArr.push(
+    //       `"${extraField}" is an extra column. Please delete the column and re-upload the file`,
+    //     ),
+    //   )
```
PikkaPikkachu, line 51: nit: needs to be reverted and same for changes below?

PikkaPikkachu (commented):
Thank you for putting this! Leaving some initial comments/suggestions and I guess you would still be looking at the failing tests!

frontend/src/features/issue/hooks/useParseCsv.tsx
ameliatay: sorry! missed this out, reverted in commit ea9c026

frontend/src/features/issue/components/BulkIssueUploadCsvCard.tsx
ameliatay: pulled out into a separate file under hooks in commit ea9c026

frontend/src/features/issue/components/BulkIssueSendLettersCard.tsx
ameliatay: felt naming it generically as `setUploadCsvErrors` might not differentiate well from row errors for new engineers looking, so renamed as `setUploadCsvGeneralAndColumnErrors` in commit ea9c026

backend/src/letters/letters.service.ts
ameliatay: done! ea9c026

backend/src/letters/letters.service.ts
ameliatay: done! ea9c026

frontend/src/features/issue/components/BulkIssueUploadCsvCard.tsx
```diff
@@ -171,12 +183,14 @@ export const BulkIssueUploadCsvCard = ({
...
-  const csvFormError = parseCsvError.length > 0
+  const csvFormError =
+    parseCsvError.length > 0 || uploadCsvGeneralAndColumnErrors.length > 0
 
   const passwordInstructionsError =
     isPasswordProtected &&
     passwordInstructions &&
-    passwordInstructions.length < 10
+    (passwordInstructions.length < MIN_PASSWORD_INSTRUCTION_LENGTH ||
+      passwordInstructions.length > MAX_PASSWORD_INSTRUCTION_LENGTH)
```
PikkaPikkachu, line 193: quick question: this seems like an accidental change?

frontend/src/features/issue/hooks/filterUploadCsvErrors.tsx
```diff
@@ -0,0 +1,31 @@
+import {
+  BulkLetterValidationColumnLevelError,
+  BulkLetterValidationError,
+  BulkLetterValidationErrorType,
+  BulkLetterValidationRowLevelError,
+} from '~shared/dtos/letters.dto'
+
+const filterUploadCsvErrors = (errors: BulkLetterValidationError[]) => {
```
PikkaPikkachu, line 8: nit: there's also another file called `useParseCsv` which handles some csv error handling logic, do you think it would make sense to put this logic there instead of creating a new file ?

shared/src/dtos/letters.dto.ts
```diff
@@ -50,6 +50,9 @@ export class CreateBulkLetterDto {
   templateId: number
   @IsDefined()
   letterParamMaps: LetterParamMaps
+  @IsDefined()
+  @IsArray()
+  letterColumnHeaders: string[]
```
PikkaPikkachu, line 55: I don't think we should be adding this field separately since `letterColumnHeaders` can be derived from `letterParamMaps` as well?

PikkaPikkachu:
Overall flow-wise makes sense. Could you help put up a small doc maybe for the new error flow — so that it's a tad bit easier to review the PR. For e.g. why we decided to add `letterColumnHeader` , and some other changes (does not have to be super comprehensive though), it will just make the review process a bit simpler :))

Also it would be nice to have some UI/error tests in the PR, so we can see how the UI looks post changes, thanks for working on this @ameliatay !

frontend/src/features/issue/components/BulkIssueUploadCsvCard.tsx
ameliatay: Is this unnecessary? The original was only `passwordInstructions.length < MIN_PASSWORD_INSTRUCTION_LENGTH` so thought the check for <560 characters was missed out, hence added it in

shared/src/dtos/letters.dto.ts
ameliatay: Was working through this logic with Keith - that there could be missing columns, so in that case the letterParamMaps will be incomplete objects and using them to check for valid headers won't suffice. Is this logic right?

frontend/src/features/issue/hooks/filterUploadCsvErrors.tsx
ameliatay: Yes! I'll shift it, wasn't sure if it was bad practice to export another function in the same file haha

ameliatay: moved in 4e3db11

ameliatay:
> ticket

Sorry! Didn't think about how I could make the PR process easier, will take note for future :-) Tried consolidating change log in [the ticket itself](https://www.notion.so/opengov/chore-Clean-up-bulk-issue-validation-errors-f42fdc991b7e41f0a5ea84fd96dd9738?pvs=4), hope this works~

ameliatay:
> Overall flow-wise makes sense. Could you help put up a small doc maybe for the new error flow — so that it's a tad bit easier to review the PR. For e.g. why we decided to add `letterColumnHeader` , and some other changes (does not have to be super comprehensive though), it will just make the review process a bit simpler :))
> 
> Also it would be nice to have some UI/error tests in the PR, so we can see how the UI looks post changes, thanks for working on this @ameliatay !

Sorry! Didn't think about how I could make the PR process easier, will take note for future :-) Tried consolidating change log in [the ticket itself](https://www.notion.so/opengov/chore-Clean-up-bulk-issue-validation-errors-f42fdc991b7e41f0a5ea84fd96dd9738?pvs=4), hope this works~

PR #245, opened 2023-08-17
fix: internal 500 error

PikkaPikkachu:
## Context

This PR resolves the 500 internal server errors we are getting. The errors are being triggered after the users are inactive for a while, due to which they either loose their previous session. However, currently our frontend might still go ahead and make calls to fetch data, [a race around condition], which results in the 500 internal server error. 

Instead we should be throwing a 401, for unauthorised access. So this PR fixes that.

## Approach

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:
[insert screenshot here]

**AFTER**:
[insert screenshot here]

## Tests

## Deploy Notes

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:

## Risks


gweiying (approved):
lgtm! thanks for debuggin this~~~

PR #246, opened 2023-08-17
test: add test cases for notifications service

PikkaPikkachu:
This PR is building on top of the previous backend testing PRs. In particular we are adding tests for the notification service. There are three test cases added :
1. Sending email notification correctly
2. Sending SMS notification correctly
3. Should throw error when notification type is incorrect.


ddanelynn (approved):
lgtm!!

PR #247, opened 2023-08-17
Feat/growthbook

gweiying:
## Context
This PR sets up the backend integration of Growthbook. We want to introduce feature flagging to our codebase so that we can release features quickly and test features out with segments of our users without rolling it out to all users at once. The frontend integration for Growthbook will be implemented separately. 

Documentation for Growthbook usage is [created on Notion here](https://www.notion.so/opengov/Guide-to-Growthbook-feature-flagging-eac4df8b642e4f0cbd2de84a57018671). 

## Approach
**FeatureFlagService**
The `FeatureFlagService` class is set up as a request-scoped NestJS injectable service. The Growthbook client is designed to be lightweight and stateless. This allows us to initiate one client per incoming request, and tag the request with identifying attributes for that user. 

**FeatureFlagModule**
The `FeatureFlagModule` is made global, so `FeatureFlagService` can imported globally without importing the individual service as a provider in every module that we want to introduce feature flagging in.

**Local development and fallbacks**
Boolean values:
In a local development environment, all features are enabled by default. 
Outside of local development, features will be disabled by default if the Growthbook client key is not provided. This is done assuming that we are more likely to be worried about accidentally releasing features that not ready (less likely for users to self-report and weirder side effects), than rolling back complete features (more likely for users to self-report).
Non-boolean values:
When querying the value of a feature flag, fallback values will be returned by default.

**Logging**
Feature key-value pairs as well as corresponding user attributes are logged on Datadog each time a feature is retrieved. This allows for easier debugging to see what flag values a user was served during their usage.

## Deploy Notes

- `letters-<ENV>-growthbook-client-key` : env var details

## Testing

The staging environment is currently deployed with `template-create` as a feature flag, please feel free to test out the feature flagging by modifying the rules for the `template-create` flag. 

If it's disabled, the [template creation page](https://staging.letters.gov.sg/admin/templates/create) should fail with the error message like below:
Screenshot 2023-08-21 at 5 23 03 PM

Otherwise, template creation should succeed. Feel free to test out segmentation rules as well. For example, the current rule only allows @open.gov.sg users to create templates


PikkaPikkachu:
Overall flow LGTM. Just one quick question (since this is only the backend), we intend to feature flag on frontend itself right? Having backend APIs fail while the frontend UI is still accessible to our users, seems like not the best UX ? For e.g. template creation page should not be made accessible to users at all, if template creation feature is being gatekeeped [but since this is only backend PR, maybe we will be addressing the remaining flagging in frontend PR ?]

backend/src/feature-flag/feature-flag.service.ts
```diff
@@ -0,0 +1,102 @@
...
+          )
+          return fallbackValue
+        },
+      } as GrowthBook
+    }
+
+    const user = this.getUserFromRequest(this.request)
+    return new GrowthBook<GrowthBookFeaturesMap>({
+      apiHost: this.config.get('growthbookApi.host'),
+      clientKey: this.config.get('growthbookApi.clientKey'),
```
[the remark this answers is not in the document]
gweiying: Makes sense, tackled here [ab7b254](https://github.com/opengovsg/letters/pull/247/commits/ab7b254faf8154917ff00eedd13fb8974a4c4177)

gweiying:
> we intend to feature flag on frontend itself right
Yep, will be tackled in a separate PR

backend/src/feature-flag/feature-flag.service.ts
```diff
@@ -0,0 +1,104 @@
...
+    @Inject(REQUEST)
+    private readonly request: Request,
+    @InjectPinoLogger(FeatureFlagService.name)
+    private readonly logger: PinoLogger,
+  ) {
+    this.growthbook = this.initializeGrowthbook()
+    this.featuresLoaded = this.loadFeatures()
+  }
+
+  private initializeGrowthbook() {
```
PikkaPikkachu, line 30: wondering if we can use our `@CurrentUser() user: User` decorator here instead of writing logic in another function `getUserFromRequest`  instead?

backend/src/feature-flag/feature-flag.service.ts
```diff
@@ -0,0 +1,104 @@
...
+        ) => {
+          console.log(
+            `Feature ${featureKey} using fallbackValue ${fallbackValue.toString()} by default as Growthbook is not initialized`,
+          )
+          return fallbackValue
+        },
+      } as GrowthBook
+    }
+
+    const user = this.getUserFromRequest(this.request)
```
PikkaPikkachu, line 58: Then we wouldn't need the logic here and the function below?

backend/src/feature-flag/feature-flag.service.ts
gweiying: Hey @PikkaPikkachu! That's a good suggestion but currently not possible. 

The request context is being passed to `FeatureFlagService` by [injecting the request object](https://docs.nestjs.com/fundamentals/injection-scopes#request-provider), which is different from how our `@CurrentUser() user: User` decorator is accessing the request object ([through the execution context](https://docs.nestjs.com/fundamentals/execution-context#executioncontext-class)). 

Since the destructuring of the request object would be quite different, I'm not sure if it would make sense to share much logic here, unless you have something in mind. If so, let's catch up offline to discuss this further?

PikkaPikkachu: Sounds good, it was more of a question and non-blocking!

PR #254, opened 2023-08-21
chore: send feedback to zendesk

ameliatay:
## Context

Currently qualitative data is stored in datadog for only 15 days, making it difficult to track and act on them. So we will be piping data to Zendesk instead. 

Once piped to Zendesk, for PM's action to review and respond in their own time, also explaining why just tracking them as Zendesk tickets suffices, over creating a new dashboard for this purpose.

Closes [this issue](https://www.notion.so/opengov/chore-Send-feedback-to-Zendesk-a10ffb08eb2c4d8881f0df8c1b002862?pvs=4)

## Approach

Use the existing mailerService to send an email to support@letters.gov.sg over deciding to use Zendesk's API to create a new ticket. Key reason being effort level is slightly higher to handle a new API and the benefits aren't proportionally rewarding. Specifically, given the simple use case mentioned under `Context`, the flexibility and control of using the API is not as necessary.

## Screenshots
Sample mail sent
Screenshot 2023-08-22 at 5 57 00 PM


backend/src/feedback/feedback.service.spec.ts
```diff
@@ -1,19 +1,34 @@
...
     info: jest.fn(),
   }
 
+  const mockConfigService = {
+    get: jest.fn(),
+  }
+
+  const mockMessageSid = 'MOCK_MESSAGE_SID'
+  const mockMailerService = {
+    sendMail: jest.fn().mockResolvedValue({ messageId: mockMessageSid }),
```
PikkaPikkachu, line 23: nit: we can probably do without these mocks, since we never call the `sendMail` in our test file? Unless you plan to write tests for checking the sending mail functionality, in which case, it would be good to keep them?

PikkaPikkachu:
Can we also update the PR with screenshots of the email body that actually got sent, you can check the emails that were received on `localhost:1080` :)) Thanks! 
Alternatively, with some setup changes, you should also be able to send emails to your actual gmail account.

backend/src/feedback/feedback.service.ts
```diff
@@ -21,6 +27,19 @@ export class FeedbackService {
       feedback_rating: rating,
       feedback_text: feedbackText,
     })
+
+    const html = `<p>User email: ${
+      currentUser.email
+    }</p><p>Datetime: ${new Date().toISOString()}</p><p>Feedback name: ${feedbackName}</p><p>Feedback rating: ${rating}</p><p>Feedback text: ${feedbackText}</p>`
+    const mail = {
+      to: 'support@letters.gov.sg',
```
PikkaPikkachu, line 35: can we put this email in a const string please? Config variable would also be fine!

backend/src/feedback/feedback.service.ts
```diff
@@ -21,6 +27,19 @@ export class FeedbackService {
       feedback_rating: rating,
```
gweiying, line 27: if you don't mind, could you also help to add a user identifier here? probably user email will be sufficient.

backend/src/feedback/feedback.service.ts
```diff
@@ -21,6 +27,19 @@ export class FeedbackService {
...
       feedback_text: feedbackText,
     })
+
+    const html = `<p>User email: ${
+      currentUser.email
+    }</p><p>Datetime: ${new Date().toISOString()}</p><p>Feedback name: ${feedbackName}</p><p>Feedback rating: ${rating}</p><p>Feedback text: ${feedbackText}</p>`
```
gweiying, line 33: seems good to me but might want to check in with Clement on a preferred format and whether it's easy for him to set rules to filter this email~

ameliatay: Clement gave the ok 👍🏻

backend/src/feedback/feedback.service.ts
ameliatay: done in e230225

backend/src/feedback/feedback.service.ts
ameliatay: extract to const string in e230225!

frontend/src/features/issue/components/BulkIssueSendLettersCard.tsx
```diff
@@ -53,6 +54,7 @@ export const BulkIssueSendLettersCard = ({
   const { getValues } = useFormContext()
   const toast = useToast({ position: 'bottom' })
   const adminAuth = useAdminAuth()
+  const hrefMailToString = `mailto:${SUPPORT_EMAIL}`
```
PikkaPikkachu, line 57: nit: can just inline this, like we did above?

PikkaPikkachu (approved):
LGTM – one tiny non-blocking nit! Let's also push to staging to test this out and get this QA-ed. 

Good to merge, once all the tests pass :)

PR #255, opened 2023-08-21
fix: snyk auto-change dependency

PikkaPikkachu:
## Context

This PR fixes the semver dependency issue here: https://app.snyk.io/org/gogovsg/project/d9d2c026-3bf1-4ba7-ac7c-485ec9499708, by moving the `autochange-log` module to dev dependency instead of dependencies. This is similar to how it's done for `engage`.

## Approach

Changed `autochange-log` to `devDependency`. 

## Testing
For testing I tried to run `npm version patch` to check if `autochange-log` is working as expected. 
I was able to create a patch version in `changelog.md`


halfwhole (approved):
lgtm, tq for fixing this!
strange that the `package-lock.json` for the frontend/backend/e2e got updated too, but looks harmless!

PR #258, opened 2023-08-22
Feat/growthbook fe

gweiying:
## Context

We have previously integrated Growthbook feature flagging on our backend in #247. We want to integrate Growthbook feature flagging on our frontend as well.

### Client key handling
We have opted to send the client key from our backend to our frontend via the `GET /config` endpoint that we have available. This shouldn't be a security risk as the client key is meant to be public anyway. Other teams have opted to [hardcode the values on the frontend](https://github.com/opengovsg/plumber/pull/124) and switch depending on environment, but since we already had the `GET /config` endpoint available for retrieving config values on the frontend, we just went with it instead to avoid having to introduce more logic on which environment key is retrieved when.

There's currently [only two config keys](https://github.com/opengovsg/letters/blob/dda4814105c666674c3d803296d9299fa93cc72a/shared/src/constants/config.ts) accessible on this endpoint, and both keys are not meant to be secret. 

### GrowthbookWrapper
`GrowthbookWrapper` contains the logic for setting up the `growthbook` client. Unlike the backend, the Growthbook client here is instantiated on client-side load, and is not request-scoped. In order to get the attribute-specific features to load, we run a `useEffect` to reset the attributes on the `growthbook` whenever the `adminUser` is changed. 

### getMockGrowthbookClient
We have also created a shared `getMockGrowthbookClient` util function to return a fallback mock Growthbook client for local development or for when Growthbook fails to be initialized (no client key etc). The behavior of the fallback client is discussed [here](https://www.notion.so/opengov/Guide-to-Growthbook-feature-flagging-on-Letters-eac4df8b642e4f0cbd2de84a57018671?pvs=4#e121885d76ff407eade1e6de404b2427)

## Testing
The staging environment is currently deployed with `template-create` as a feature flag, please feel free to test out the feature flagging by modifying the rules for the `template-create` flag.

If you've logged in with an email ending with `@open.gov.sg`, you should be able to see the Create Template button like so, 
Screenshot 2023-08-24 at 6 11 09 PM

Otherwise, if you're logged in with an email not ending with `@open.gov.sg`, for example `support@letters.gov.sg`, you should instead see the `Request New Template` button like so,
Screenshot 2023-08-24 at 6 12 45 PM


frontend/src/features/feature-flag/GrowthbookWrapper.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const { growthbookApiKey } = useGrowthbookApiKey()
+  const { adminUser } = useAdminUser()
+
+  const growthbook = useMemo(() => {
+    if (process.env.NODE_ENV === 'development' || !growthbookApiKey)
+      return getMockGrowthbookClient({
+        isDevEnv: process.env.NODE_ENV === 'development',
+      }) as GrowthBook
+    return new GrowthBook<GrowthBookFeaturesMap>({
+      apiHost: 'https://cdn.growthbook.io',
```
PikkaPikkachu, line 22: I remember this was also part of config, shouldn't we be getting this from config as well, similar to how we get the API key?
(Just for future maintainability) — having the const in common place might be better? [non-blocking]

frontend/src/features/dashboard/TemplatesPage.tsx
```diff
@@ -1,20 +1,39 @@
...
+            marginLeft={8}
+            alignSelf="start"
+            onClick={() => navigate('/admin/templates/create')}
+          >
+            Create New Template
           </Button>
-        </Link>
+        ) : (
+          <Link
+            href="https://go.gov.sg/lettersg-onboarding-new-template"
```
PikkaPikkachu, line 29: nit: maybe we can put this link const in `links.ts` file to avoid constants inside components :)

frontend/src/features/create/CreateTemplatePage.tsx
```diff
@@ -1,19 +1,25 @@
...
 import { CreateTemplateModal } from './components/CreateTemplateModal'
 
 export const CreateTemplatePage = (): JSX.Element => {
   const navigate = useNavigate()
   const [templateContent, setTemplateContent] = useState('')
   const { isOpen, onOpen, onClose } = useDisclosure()
+  const isTemplateCreateOn = useFeatureIsOn(
+    GrowthBookFeatures.templateCreate as string,
+  )
+  if (!isTemplateCreateOn) return <Navigate to={`/admin/letters`} />
```
PikkaPikkachu, line 22: nit: I have seen in some places, we are doing something like so:
```
${routes.admin.index}/${routes.admin.letters}
```
which might be neater (also good for long run -- we wouldn't have to manually change within code if we decide to rename our routes :)) )

same for the routing below ?

frontend/src/features/dashboard/TemplatesPage.tsx
```diff
@@ -1,20 +1,39 @@
...
+            isExternal
+          >
+            <Button marginLeft={8} alignSelf="start">
+              Request New Template
+            </Button>
+          </Link>
+        )}
```
PikkaPikkachu, line 36: nit: since the button styling here and everything else kind of remains the same, I was wondering, if we could actually do something like this:
```
  <Button
          marginLeft={8}
          alignSelf="start"
          onClick={() =>
            isTemplateCreateOn
              ? navigate('/admin/templates/create')
              : window.open(
                  'https://go.gov.sg/lettersg-onboarding-new-template',
                )
          }
        >
          {isTemplateCreateOn ? `Create New Template` : `Request New Template`}
        </Button>
```

to avoid maintaining duplicate stylings  — so that we are not making changes in two places next time we change button styles? But up to you!

PikkaPikkachu (approved):
LGTM ! There are some nits and suggestions on use of constants and tiny refactoring, that can be handled in separate tasks or this one!

frontend/src/features/feature-flag/GrowthbookWrapper.tsx
gweiying: I was wondering whether it needs to be something we fetch via the API call or whether we should just hardcode it. Since it's unlikely to change, I'd prefer not fetching it, but I'm also happy to move it out into a const otherwise! We can discuss this offline?

frontend/src/features/create/CreateTemplatePage.tsx
gweiying: [7accf7d](https://github.com/opengovsg/letters/pull/258/commits/7accf7d094bd29de3cf9994011b6b0385e61bdc2)

frontend/src/features/dashboard/TemplatesPage.tsx
gweiying: [7accf7d](https://github.com/opengovsg/letters/pull/258/commits/7accf7d094bd29de3cf9994011b6b0385e61bdc2)

frontend/src/features/dashboard/TemplatesPage.tsx
gweiying: Good idea! Adopted during the merge commit [0f05836](https://github.com/opengovsg/letters/pull/258/commits/0f05836d0fed020c5b8c937643101c16cb85c4f4)

PR #259, opened 2023-08-23
chore: increase nestjs request entity json accepted size

gweiying:
## Context

Clement had difficulty uploading the Customs template on staging. When testing on local, the request failed with the error `Nest.js - request entity too large PayloadTooLargeError: request entity too large`.

## Approach

Explicitly sets the accepted NestJS request JSON size to 20MB
See https://github.com/nestjs/nest/issues/529#issuecomment-376576929

## Test
- [ ] Test on staging


PikkaPikkachu (approved):
Nice! Just wondering if we should store the size limit in a environment variable to make it configurable in production, without having to release 😅 ?

gweiying:
> Nice! Just wondering if we should store the size limit in a environment variable to make it configurable in production, without having to release 😅 ?

I think let's keep it simple for now and monitor? If we need to change it frequently then I think your suggestion makes sense~

PR #260, opened 2023-08-23
chore: admin query keys

disKeith:
## Context

This PR aims to replace the existing hardcoded query keys with a defined object to provide:
 - Centralised definition of query keys for enhanced maintainability
 - Consistency in query key naming and structure
 - Easier reusability without inconsistencies or errors
 - Type safety, hints, and autocompletion

ref: https://github.com/opengovsg/letters/pull/239#discussion_r1294087801

Closes [Notion Task](https://www.notion.so/opengov/chore-Use-object-types-to-manage-React-query-keys-e85c0a02452740bebd9bebc2d9b4b1fa?pvs=4)

## Approach

1. Implement `adminQueryKeys` object to manage all admin-related query keys
2. Update existing admin queries to use the new object


frontend/src/constants/query-keys.ts
```diff
@@ -3,3 +3,15 @@ export const publicQueryKeys = {
   letters: (letterPublicId: string, password?: string) =>
     [publicQueryKeys.base, letterPublicId, password] as const,
 }
+
+export const adminQueryKeys = {
+  who_am_i: 'who-am-i' as const,
+  tinymce: 'tinymce-api-key' as const,
+  templates: 'templates' as const,
+  template: (templateId: number) =>
+    [adminQueryKeys.template, templateId] as const,
```
gweiying, line 12: what do you think of something like the following instead:
```
export const adminQueryKeys = {
  templates: (templateId?: number) => templateId ? ['templates', templateId] : ['templates']
}
```

gweiying (changes requested):
I found an article [here](https://tkdodo.eu/blog/effective-react-query-keys#structure) that might actually help inform our `queryKey` structuring abit better!
 
Thinking that the approach suggested in the blog above would be quite useful especially for batches, because we actually have batches - pagination and batches - item view.

frontend/src/constants/query-keys.ts
disKeith: thanks, that looks better - updated in https://github.com/opengovsg/letters/pull/260/commits/2b6e177fbd3c81f8542dccc2d0dcc852e98c3c84 with new structure

frontend/src/constants/query-keys.ts
```diff
@@ -7,11 +7,16 @@ export const publicQueryKeys = {
...
-    [adminQueryKeys.template, templateId] as const,
-  batches: 'batches' as const,
-  batch: (batchId: number) => [adminQueryKeys.batches, batchId] as const,
-  batch_view: (limit: number, offset: number) =>
-    [adminQueryKeys.batches, limit, offset] as const,
+  templates: (templateId?: number) =>
+    templateId
+      ? (['templates', templateId] as const)
+      : (['templates'] as const),
+  batches: (type?: 'detail' | 'paginate', idOrOffset?: number) => {
```
gweiying, line 14: Small suggestion. 

For the pagination, we will need 2 values, limit and offset. Thinking that we possibly accept an object here instead, for example`batches: ({ batchId, limit, offset })`, with a switch case on whether `limit` & `offset` are defined, or whether `batchId` is defined, or none of the above? This would mean that for a person using the function, they could just call it simply by passing `adminQueryKeys.batches({limit: 10, offset:10 })` for example without needing to look up what `type` refers to.

disKeith: roger, updated in https://github.com/opengovsg/letters/pull/260/commits/f980729c10f1e06b7c7ed58dcb69f6b53d4cac36

frontend/src/constants/query-keys.ts
```diff
@@ -3,3 +3,30 @@ export const publicQueryKeys = {
   letters: (letterPublicId: string, password?: string) =>
     [publicQueryKeys.base, letterPublicId, password] as const,
 }
+
+export const adminQueryKeys = {
+  who_am_i: 'who-am-i' as const,
```
gweiying, line 8: sorry, final nit from me! can we use camelCase here?

frontend/src/constants/query-keys.ts
```diff
@@ -3,3 +3,30 @@ export const publicQueryKeys = {
...
+      : (['templates'] as const),
+  batches: ({
+    batchId,
+    limit,
+    offset,
+  }: {
+    batchId?: number
+    limit?: number
+    offset?: number
+  }) => {
```
gweiying, line 22: is it also possible to make this whole object optional? that way if no object is passed e.g. `batches()` it would directly return `['batches']`

disKeith: addressed comments in https://github.com/opengovsg/letters/pull/260/commits/5a22949238237db312d0031513aa48ae7c947282 !

gweiying (approved):
Overall looks good, can we push to staging to test out and verify that all queries are being cached/ invalidated correctly before merging? Can try React Query dev mode tool to help with profiling this

PR #262, opened 2023-08-24
Chore/feedback to zendesk

ameliatay:
## Context

Currently qualitative data is stored in datadog for only 15 days, making it difficult to track and act on them. So we will be piping data to Zendesk instead. 

Once piped to Zendesk, for PM's action to review and respond in their own time, also explaining why just tracking them as Zendesk tickets suffices, over creating a new dashboard for this purpose.

Closes [this issue](https://www.notion.so/opengov/chore-Send-feedback-to-Zendesk-a10ffb08eb2c4d8881f0df8c1b002862?pvs=4)

## Approach

Use the existing mailerService to send an email to support@letters.gov.sg over deciding to use Zendesk's API to create a new ticket. Key reason being effort level is slightly higher to handle a new API and the benefits aren't proportionally rewarding. Specifically, given the simple use case mentioned under `Context`, the flexibility and control of using the API is not as necessary.

## Screenshots
Sample mail sent
Screenshot 2023-08-22 at 5 57 00 PM


backend/src/feedback/feedback.service.ts
```diff
@@ -1,26 +1,48 @@
...
     const { feedbackName, rating, feedbackText } = feedbackDto
+    const email = currentUser.email
+    const datetime = new Date().toISOString()
 
     this.logger.info({
       // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
       msg: `Feedback on ${feedbackName} (Rating: ${rating}): ${feedbackText}`,
       // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+      user_email: email,
+      date_time: datetime,
```
gweiying, line 30: Haha tiny nit & non-blocking - actually we don't need to log `date_time` here cause all logs are registered automatically with `date_time`.

gweiying (approved):
lgtm!

backend/src/feedback/feedback.service.ts
```diff
@@ -1,26 +1,48 @@
...
       feedback_name: feedbackName,
       feedback_rating: rating,
       feedback_text: feedbackText,
     })
+
+    const html = `<p>User email: ${email}</p><p>Datetime: ${datetime}</p><p>Feedback name: ${feedbackName}</p><p>Feedback rating: ${rating}</p><p>Feedback text: ${feedbackText}</p>`
```
gweiying, line 36: Please help with adding a variable here to help Clement filter whether the feedback is received on staging or production~

ameliatay: addressed in ad8573~

PR #263, opened 2023-08-24
Feat/announcement banner

gweiying:
## Context

Removes the existing application-wide beta banner, and adds a feature-flagged officer-side announcement banner. The banner contents can now be set on [Growthbook](https://app.growthbook.io/features/announcement-banner), which means that our product folks will be able to easily change the banner contents whenever they need e.g. in the case of any email or SMS slowdowns or new announcements.

### Future work
For now, the banner is not closable. 

To make the banner closable, we would need to use a custom close button so that we can fire off an event to save the viewed message in the local storage. However, the design-system version that we're using unfortunately [does not have a working custom button](https://github.com/opengovsg/design-system/pull/399), so we need to either:
a) implement our own banner
b) upgrade our design system to v1.3 and above ([breaking changes on a couple of components](https://github.com/opengovsg/design-system/pull/324)) 

Personally I think it'd be worth tackling the breaking changes to upgrade our design system since we want to enjoy the benefits of a maintained design system, but I think we can tackle this in a separate piece of work later (non-urgent for now).

## Approach

- Removes the existing app-wide banner
- Introduces a feature-flagged officer-side announcement banner (supports markdown so links can be used in announcement banner too)

## Screenshots
**BEFORE**
Screenshot 2023-08-25 at 6 32 04 PM
Screenshot 2023-08-25 at 6 34 09 PM

**AFTER**
General announcement banner (officer side)
Screenshot 2023-08-25 at 6 31 34 PM

Custom announcement banner (officer side)
Screenshot 2023-08-25 at 6 29 35 PM

No banner on public view
Screenshot 2023-08-25 at 6 32 42 PM


PikkaPikkachu (approved):

disKeith (approved):
lgtm!

PikkaPikkachu:
Thanks for fixing tests 😅 !

PR #267, opened 2023-08-29
Fix/button sizes

gweiying:
## Context

Three minor UI fixes:
- standardizes button sizes across the whole application (previously was mix of 'large', 'sm'), now defaults to 'md' which is OGP-default)
- fixes top and left alignment between the Issued Letters page and the Templates page
- replaces single appearance of `<Text>Loading</Text>` with `<Spinner/>` component

### Buttons
**BEFORE**:
Different sized buttons
Screenshot 2023-08-29 at 9 25 32 AM
Screenshot 2023-08-29 at 9 25 57 AM
Screenshot 2023-08-29 at 9 24 33 AM
Screenshot 2023-08-29 at 9 20 59 AM

**AFTER**:
Standardized buttons
Screenshot 2023-08-29 at 9 22 38 AM
Screenshot 2023-08-29 at 9 23 12 AM
Screenshot 2023-08-29 at 9 22 28 AM
Screenshot 2023-08-29 at 9 23 00 AM

### Alignment
**BEFORE**
Screen Recording 2023-08-29 at 9 30 21 AM

**AFTER**
Screen Recording 2023-08-29 at 9 28 56 AM


frontend/src/features/dashboard/components/TemplatesBody.tsx
```diff
@@ -20,7 +20,7 @@ export const TemplatesBody = () => {
             </WrapItem>
           ))
         ) : (
-          <Text>Loading</Text>
+          <Spinner />
```
PikkaPikkachu, line 23: we have a component `<Loading/>` in our code base — we can also re-use that, if you'd like? But since you didn't have to add much styling here, not sure if it should be useful, I think that would automatically centralise the spinner for you :)

PikkaPikkachu (approved):
Looks nice! :D

frontend/src/features/dashboard/components/TemplatesBody.tsx
gweiying: Hm we seem to be using these components interchangably throughout the codebase! Happy to take a future pass through the codebase to standardize (all Loading or all Spinner), but deprioritising for now.

PR #268, opened 2023-08-29
chore: email copy changes

PikkaPikkachu:
## Context

This PR makes some copy changes to the existing email copy as proposed [here](https://www.notion.so/opengov/Email-copy-changes-20100594f2cc4b34912f434b96239ff9#4e6c458a80c64cb3a15892b784a23e52)

## Approach

1. extracted logic for email copy in shared function
2. there were some complexities in calling this logic in frontend, so decided that we can leave this for future work for now. 

## Before & After Screenshots

**BEFORE**:
[insert screenshot here]

**AFTER**:
[insert screenshot here]


backend/src/notifications/email-notifications.service.ts
```diff
@@ -94,46 +92,34 @@ export class EmailNotificationService {
...
     }[]
   > {
     const user = await this.usersRepository.findOne({
       where: { id: userId },
       select: ['email'],
     })
+
+    const templateId = letters[0].templateId
+    const template = await this.templateRepository.findOneBy({ id: templateId })
+
```
PikkaPikkachu, line 109: had to go with this hacky approach for now to fetch the template name :/

gweiying: Seems ok but agree that it's not ideal. Let's open a ticket to clean up later!

gweiying (approved):
lgtm!

shared/src/util/emailTemplate.ts
```diff
@@ -0,0 +1,23 @@
...
+export const getEmailTemplateBody = (
+  officerEmail: string,
+  letterPublicId: string,
+  domainName: string,
+): string => {
+  return `<p> <i>This communication is legitimately issued via LetterSG, the official platform for trusted e-letters from the Singapore Government. </i></p>
+  
+  <br/>
+        <p></p>
+        <p >Dear Sir/Madam,</p>
```
disKeith, line 12: awkward spacing here 😂

shared/src/util/emailTemplate.ts
```diff
@@ -0,0 +1,23 @@
...
+        <p></p>
+        <p></p>
+        <p >You have been issued a letter by <a target="_blank" href="mailto:${officerEmail}">${officerEmail}</a>.</p>
```
disKeith, line 15: and here

PR #269, opened 2023-08-29
Chore/remove download page

gweiying:
## Context

This PR removes the download page at the end of the letter issuance flow. Instead, the user is directly sent to the Issued Letters dashboard on completion (and after submitting the feedback modal) to download the CSV.

**Future work**
We have discussed that the users should be prompted when their new batch appears in the batches table, possibly through a `New` badge or something in the new table row. However, we have not implemented this for now.

This PR also temporarily disables the E2E test for gnerated password protected letters to contain the matched params in the uploaded CSV. The CSV downloaded from the batches table do not contain the params of the letters as the params are stored encrypted, hence we are unable to use the CSV download from the batches table to check that the generated letters contain the params. We should rewrite the tests to use the params from the uploaded CSV to check against the generated letters instead.

## Before & After Screenshots

**BEFORE**:
Screen Recording 2023-08-29 at 1 43 17 PM

**AFTER**:
Screen Recording 2023-08-29 at 1 42 23 PM

## Tests

## Deploy Notes

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:

## Risks


gweiying:
Closing as this is irrelevant in the newest designs

PR #270, opened 2023-08-29
fix: feedback auth issue

PikkaPikkachu:
## Context

While submitting feedback from citizen side we realised that the feedback endpoints need authentication for anyone to submit feedback. 
Removing the need for authentication to submit feedback. 

## Approach

1. Make feedback a public endpoint

**Bug Fixes**:

- Remove authguard and use `@OptionalUser` instead of `@CurrentUser` for submitting feedback.

## Before & After Screenshots

https://github.com/opengovsg/letters/assets/25703976/ae3597c1-1dc0-4068-97db-406ae5749438

## Future Work
1. We probably should add validation before submitting feedback, which is currently missing.


ameliatay (approved):

gweiying (approved):
lgtm! thanks for the quick change. in the future we might want to allow MOPs to provide emails so they can be contacted but can deprioritise for now

PR #276, opened 2023-08-31
Fix/password protect ff

gweiying:
## Context

This PR feature-flags and disables the generation of password-protected letters. We will release the feature at a later date after assessing our internal SMS policy on sending links with password-protected letters.

## Approach

- Adds a new feature flag to Growthbook `'password-protect'`. Feature flag is currently disabled on all environments
- Both frontend and backend are affected by the feature flag. When disabled, the changes are as follows
  - Frontend: "Are letters confidential?" section removed
  - Backend: Exception thrown if bulk create or bulk validate request is submitted with passwords or password instruction. We have no single create or single validate endpoints for now, so this should suffice.
- Unit tests added on backend controller endpoint to verify that passwordProtected letters are not created and an exception is thrown if password protect is not enabled. 

## Before & After Screenshots

**BEFORE**:
Screenshot 2023-08-31 at 8 40 58 PM

**AFTER**:
Screenshot 2023-08-31 at 8 41 44 PM

## Tests
Test plan: Verify that disabling the feature works as expected
- [x] For frontend: "Are letters confidential?" section should be removed
- [x] For backend: Make the following Postman requests
  - [x] bulk create without passwords or password instructions: should succeed as per normal
  - [x] bulk create with passwords or password instructions: should fail with error "Feature is not enabled"
  - [x] bulk validate without passwords or password instructions: should succeed as per normal
  - [x] bulk validate with passwords or password instructions: should fail with error "Feature is not enabled"

## Deploy Notes

Note that new feature flag on Growthbook `'password-protect'` is disabled by default on all environments, so we do not need to make any further changes during deploy.


backend/src/letters/letters.controller.ts
```diff
@@ -35,6 +39,17 @@ export class LettersController {
...
     if (!template) throw new NotFoundException('template not found')
+
+    const isPasswordProtectEnabled = await this.featureFlagService.isOn(
+      GrowthBookFeatures.passwordProtect,
+    )
+    const containsPasswordProtectParams =
+      (bulkRequest.passwords && bulkRequest.passwords.length > 0) ||
+      !!bulkRequest.passwordInstructions
+    if (!isPasswordProtectEnabled && containsPasswordProtectParams) {
+      throw new BadRequestException('Feature not enabled')
```
thanhdatle, line 50: nit: password protected feature not enabled

backend/src/letters/letters.controller.ts
```diff
@@ -35,6 +39,17 @@ export class LettersController {
     const { templateId, createNotificationDto } = bulkRequest
     const template = await this.templatesService.findOne(templateId)
     if (!template) throw new NotFoundException('template not found')
+
+    const isPasswordProtectEnabled = await this.featureFlagService.isOn(
```
thanhdatle, line 43: ideally, the check should be done at service layer (lettersService.bulkCreate) since the function could have been called from other services/controllers
But this should be fine then since this is the only place where we call bulkCreate

gweiying: I thought to do the check at the controller layer because was easier to see that there's no other existing endpoints that use this actually

backend/src/letters/letters.controller.ts
gweiying: Do we want to expose this? 

I was planning to leave all error message  for disabled feature flags generic, so that we don't expose what features are disabled. Though, our codebase is public anyway so this argument isn't particularly strong... I don't feel too strongly about it, so it'd be up to you.

backend/src/letters/letters.controller.ts
```diff
@@ -59,6 +74,16 @@ export class LettersController {
     const { templateId } = bulkRequest
     const template = await this.templatesService.findOne(templateId)
     if (!template) throw new NotFoundException('template not found')
+
+    const isPasswordProtectEnabled = await this.featureFlagService.isOn(
+      GrowthBookFeatures.passwordProtect,
+    )
+    const containsPasswordProtectParams =
```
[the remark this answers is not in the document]
thanhdatle: and we will need tests particularly for this function as well

thanhdatle (approved):

PR #279, opened 2023-09-04
fix: remove get help from profile drop down menu

ameliatay:
## Context

For visibility and a better user experience, a "Get Help" button was added to the top navbar, duplicating the same button already existing in the profile drop down. As such, we are removing it from the profile drop down.

Closes [this issue](https://www.notion.so/opengov/bug-Remove-get-help-from-profile-drop-down-dc3a47f5d1b242f5ad8cdcb4378fdae3?pvs=4)

## Approach

Remove "Get Help" from MenuList in UserMenu.tsx

## Before & After Screenshots

**BEFORE**:
Screenshot 2023-09-04 at 10 57 53 AM

**AFTER**:
Screenshot 2023-09-04 at 10 57 17 AM


PikkaPikkachu (approved):
LGTM!

PR #280, opened 2023-09-04
fix: slide component covering navfooter links

ameliatay:
## Context

When introducing "Help us improve" feedback button in citizen view UI, I used a slide component and set the width to 50% of the screen, resultantly covering the buttons like "Contact Us" "Guide" etc in the nav footer.

While debugging, also came across another small UI issue where the feedback modal does not fit into the frame for certain phone screen sizes.

Closes [this issue](https://www.notion.so/opengov/bug-Broken-contact-us-links-on-Public-Letter-page-079533684ae4488e9ddf8af4247fe199?pvs=4)

## Approach

1. Reduce height of chakra slide to 0 so it takes up no screen space (Width remains at 20% but 0 * 20% = 0 area)
2. Reduce width of feedback modal to 98% and align it right to ensure modal always fits into screen for all screen sizes

## Before & After Screenshots

**BEFORE**:
https://github.com/opengovsg/letters/assets/83273830/8540332b-9159-40e6-be60-b6ef656c830d

**AFTER**:
https://github.com/opengovsg/letters/assets/83273830/54173941-831d-4a34-8432-70b170bd1e36


frontend/src/features/public/components/FeedbackButton.tsx
```diff
@@ -12,7 +12,11 @@ export const FeedbackButton = ({
   onToggle,
 }: FeedbackButtonProps) => {
   return (
-    <Slide direction="right" in={!isOpen} style={{ width: '50%' }}>
+    <Slide
+      direction="right"
+      in={!isOpen}
+      style={{ width: isDesktop ? '10%' : '20%' }}
```
PikkaPikkachu, line 18: I think making this just 10% should work for both ?

ameliatay: reducing width to 10% will affect button look on smaller screen sizes!

gweiying (approved):
fix seems to work well for me! thanks for the good find w 0 height

PikkaPikkachu (approved):
Same here, tested on staging, thanks :D

PR #283, opened 2023-09-07
feat: add previewing ability

PikkaPikkachu:
## Context

During template creation, the only way for me to preview my created template is to save it locally and then preview in the `All templates` sections. Sometimes formatting looks strange on the final version and I have to go back and re-edit. 

Adding a simple `Preview Template` button to make template editing easier and prevent over-populating the staging DB with testing templates that get saved without a preview first. 

Kind of open to hear thoughts on this PR from @disKeith  and @gweiying too :))

NOTE: This is for our INTERNAL team ONLY, in hopes of easing the creation work. Also hoping to solve the problem of direct DB edits 😓 

## Approach

1. Added a preview modal. 

## Before & After Screenshots

**BEFORE**:
Screenshot 2023-09-07 at 6 26 32 PM

**AFTER**:

https://github.com/opengovsg/letters/assets/25703976/767b27a7-9e2c-4f50-a348-9c9d06f35154

https://github.com/opengovsg/letters/assets/25703976/9555b6c7-1820-4114-b3c1-0b8be3d766b9

## Future Work:
1. We can extend this previewing to adding field names in the template and on the spot adding of test fields to check for the generated letter looks like.
2. We should probably give OGP folks edit access to templates, so that we don't have to constantly delete or edit directly in staging DB [now that we have feature flags, it should be easier!]


gweiying:
Big fan of the idea~ Thanks for proactively adding it I've also thought about adding it previously but you've beaten me to it!

Will review abit later!

frontend/src/features/create/PreviewTemplateModal.tsx
```diff
@@ -0,0 +1,76 @@
...
+  onClose: () => void
+  templateContent: string
+}
+
+export const PreviewTemplateModal = ({
+  isOpen,
+  onClose,
+  templateContent,
+}: PreviewTemplateModalProps): JSX.Element => {
+  const transformScale = calculateTransformScale(DRAWER_WIDTH_IN_PIXELS)
```
gweiying, line 31: Not sure why `DRAWER_WIDTH_IN_PIXELS` seems to work well here as a default width. 

I guess if it's working we could keep this temporarily, but not sure if it was intentionally kept! Maybe can leave a comment for future cleanup if so, since this is only an internal function right now~

gweiying (approved):
awesome stuff!

one small question about `DRAWER_WIDTH_IN_PIXELS` but not blocking~

frontend/src/features/create/PreviewTemplateModal.tsx
PikkaPikkachu: Yeah I just used it here, because it seemed to work well. Tried other variations, actually didn't work this well haha. Yup probably good to add the comment as well!

PR #284, opened 2023-09-11
chore: add columns to notifications table

PikkaPikkachu:
## Context

Closes [insert issue #]

## Approach

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:
[insert screenshot here]

**AFTER**:
[insert screenshot here]

## Tests

## Deploy Notes

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:

## Risks


backend/src/database/entities/notification.entity.ts
```diff
@@ -32,6 +34,16 @@ export class Notification {
...
 
+  @Column('text', { nullable: true })
+  errorMessage: string
+
+  @Column({
+    type: 'enum',
+    enum: NotificationStatus,
+    default: NotificationStatus.PENDING,
+  })
+  status: string
```
disKeith, line 45: small question here: would assigning the column to be of type `NotificationStatus` be more apt?

disKeith (approved):
lgtm apart from the small question - regarding the enum for emails that are not whitelisted, my preliminary thoughts on it is that having a separate type to represent that the email was blocked/denied from notifications might be useful (seems easier to trace/handle)! since the respective statuses you implemented wasn't intended to represent that scenario :)

shared/src/dtos/notifications.dto.ts
```diff
@@ -0,0 +1,5 @@
+export enum NotificationStatus {
+  SUCCESS = 'success',
```
gweiying, line 2: Can we call this "SENT" instead of "SUCCESS" so that in the future we can add more notification types for the deliverability of the letter?

backend/src/notifications/email-notifications.service.ts
```diff
@@ -44,6 +45,8 @@ export class EmailNotificationService {
     await Promise.all(
       templatedMessages.map(
         async ({ message, letterId, templateName }, index) => {
+          let notificationStatus = NotificationStatus.PENDING
+          let errorMessage = undefined
           const recipient = recipients[index]
           const isEmailAllowListed = this.isInEmailAllowlist(recipient)
           let providerId = `DEV_MODE_NO_EMAIL_SENT_${randomUUID().substring(
```
gweiying, line 52: Could we also only set this explicitly for log only messages instead of letting the message id in line 75, 76 override it?

It was abit confusing to see this as a provider id in production for the recent error case

backend/src/notifications/email-notifications.service.ts
```diff
@@ -70,17 +73,25 @@ export class EmailNotificationService {
...
               // (for e.g <xxx-xxx-xxx@mail.postman.gov.sg>)
               providerId = messageId as string
+              notificationStatus = NotificationStatus.SUCCESS
               this.logger.info(`Sending mail to ${recipient}`)
             } catch (error) {
+              if (error instanceof Error) {
+                errorMessage = error.message
+              }
+              notificationStatus = NotificationStatus.FAIL
               this.logger.error(error)
```
gweiying, line 83: For this and the following SMS notification service, can we make the error message more informative as well, for example `Failed to send message to ${receipient}, error: ${error}`

backend/src/notifications/sms-notifications.service.ts
```diff
@@ -70,22 +71,35 @@ export class SmsNotificationsService {
...
             logOnly,
           )
+          // we mark the un-allowlisted emails to be in pending state,
+          // since they have not been sent
+          notificationStatus = logOnly
+            ? NotificationStatus.PENDING
+            : NotificationStatus.SUCCESS
         } catch (error) {
+          if (error instanceof Error) {
+            errorMessage = error.message
```
gweiying, line 89: Concerned whether this error message will be easy to index in the future, if say we want to only select notifications that had 429 error codes in the last two days for retrying - a string error message will be slower to index. Wondering whether it'd be worth having an error type as well as an error message - Do you mind looking at other repos such as Postman to see how they store errors received from the provider?

backend/src/notifications/sms-notifications.service.ts
```diff
@@ -70,22 +71,35 @@ export class SmsNotificationsService {
...
         let messageSid
+        let notificationStatus = NotificationStatus.PENDING
+        let errorMessage = undefined
         try {
           messageSid = await this.smsClient.sendMessage(
             message,
             standardizedRecipient,
             logOnly,
           )
+          // we mark the un-allowlisted emails to be in pending state,
```
gweiying, line 82: might have meant sms here!

backend/src/database/entities/notification.entity.ts
gweiying: yep adding on, i think what Keith meant here is to change from `status: string` to `status: NotificationStatus` in line 45.

PikkaPikkachu: Makes sense haha

shared/src/dtos/notifications.dto.ts
PikkaPikkachu: Aahhh! Oooopss! Thanks for sharing again 😓

backend/src/notifications/sms-notifications.service.ts
PikkaPikkachu: As discussed offline, we will be recording the `Invalid recipient` from the postman error message if the email is blacklisted.

shared/src/dtos/notifications.dto.ts
```diff
@@ -0,0 +1,6 @@
+export enum NotificationStatus {
+  SENT = 'sent',
+  FAIL = 'fail',
+  PENDING = 'pending',
+  INVALID_RECIPEINT = 'invalid recipient',
```
gweiying, line 5: minor nit here, could we make the enums capitalized and joined by underscore, similar to "EMAIL" and "TWILIO"  for example?

gweiying (approved):
overall looks sound to me! thanks, just left one minor nit inline

gweiying:
Noting on the above open ended questions:
> Should default status be pending — this would also mean that when we run this migration, we will set previous notifications to pending state, we can manually override by setting them all to success.

We've decided to only mark the successful ones as sent, and the remainder will be left as pending

> When the email or phone number is not allow listed, we are setting notification to pending, which makes sense because the notification was never sent out. We can create another state in the enum, which is Not Allowlisted to handle this case — open to hearing thoughts on this as well.

Seems fine for now!

PR #286, opened 2023-09-12
chore: upgrade design system to 1.6

PikkaPikkachu:
## Context

Upgrading the design system to v1.6. Doing it before we kick-off the revamp items. 

## Approach

1. Just updated the design system
2. Resolved one breaking change. 
3. Did some basic testing.

## Tests

E2E tests pass.


frontend/src/components/Banner/OfficerAnnouncementBanner.tsx
```diff
@@ -11,7 +11,7 @@ export const OfficerAnnouncementBanner = () => {
   )
   if (!announcementText) return <></>
   return (
-    <Banner variant="info" isDismissable={false} useMarkdown>
+    <Banner variant="info" isDismissable={false}>
```
PikkaPikkachu, line 14: I wasn't sure here, but I don't think markdown is important here ?

gweiying: Would be very useful to support links in announcement banner actually. Are there instructions on how to use markdown with `Banner`?

PikkaPikkachu: As discussed over slack, this [PR](https://github.com/opengovsg/design-system/pull/324) removes the `useMarkdown`, the expectation is that we should be able to support it on our end. Since this is not too urgent, we have decided to create a follow-up task for this. Task [here](https://www.notion.so/opengov/2ccadbea5c734aa09614b4ca8cc7177e?v=76cff9fca57b496f857c990e2639af2f&p=4b71713849ae4dbcb870e255beae300f&pm=s)

gweiying (approved):
lgtm! thanks for tackling this and opening the follow task

PR #288, opened 2023-09-14
feat: upload csv string preprocesing for email and phone numbers

ameliatay:
## Context

A user gave feedback that it was incredibly troublesome to strip the leading and trailing whitespaces from emails in order to pass the validation.

Closes [this issue](https://www.notion.so/opengov/CSV-Upload-Input-string-pre-processing-for-emails-phone-numbers-c5b08de8ba5744eaa8e5d176ba73e0ed?pvs=4)

## Approach

We are using Papa Parse CSV parser to process our csv uploads from users. Papa Parse 5 has 2 optional configs in parse() that allow us to perform transformations on the values within the file before returning processed objects.

1. transform - Function that takes in (value, header) where value is the value of the "cell" in the CSV file and header is the corresponding column that it belongs to (if header = false, index of column is returned instead) and the return value replaces the original "cell" value
2. transformHeader - Only available when header = true - Function that takes in (header, index) and the return value replaces the header

In this case we could perform string processing on not just emails and phone numbers provided, but all fields including headers. However, this PR only implemented preprocessing for email and phone number fields for the following reasons.

1. Header fields - Does not require users to make big / troublesome changes in the event there are blocking errors that require them to remove whitespaces
2. Other fields - Possibility of intentional leading and trailing whitespaces

## Before & After Screenshots

**Email validation**

Input CSV
Screenshot 2023-09-14 at 4 25 05 PM

Before
Screenshot 2023-09-14 at 4 26 35 PM

After
Screenshot 2023-09-14 at 4 27 04 PM

**Phone Number validation**

Input CSV
Screenshot 2023-09-14 at 4 27 50 PM

Before
Screenshot 2023-09-14 at 4 28 42 PM

After
Screenshot 2023-09-14 at 4 29 07 PM

## Tests

- Amend existing E2E test to take in uploaded CSV files where email / sms fields have leading / trailing whitespaces and verify that they successfully pass the API's validation
- This assumes the API's validation will never allow invalid email / sms values to go through (Invalid = Have leading / trailing whitespaces)


PikkaPikkachu (commented):
Wooooh, thanks for the super elegant solution on this! 
We should just add some E2E tests before merging for testing this change, rest looks great!

ameliatay:
> Wooooh, thanks for the super elegant solution on this! We should just add some E2E tests before merging for this change, rest looks great!

@PikkaPikkachu is it alright to solely amend existing e2e to check if email / sms fields with leading and trailing whitespaces are still successfully processed? this assumes that the API's validation will never allow email / sms fields to go through if not trimmed which i think is a fair assumption so we'll just need to check if uploaded CSVs with untrimmed email / sms successfully goes through

a more stringent e2e check would be scanning the contents of the completedCsv file but not sure if that's necessary

PR #289, opened 2023-09-15
chore: stashing changes

gweiying:
## Context

Closes [insert issue #]

## Approach

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:
[insert screenshot here]

**AFTER**:
[insert screenshot here]

## Tests

## Deploy Notes

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:

## Risks


gweiying:
> Thanks @gweiying for walking me through the code changes just now! Just noting down items here from call:
> 
>     1. Address any remaining TODOs (no pressure haha)
> 
>     2. Try to feature-flag IFF not too much effort (for safer launch) [not high pri]
> 
>     3. Comment out failing E2E tests for now to unblock
> 
>     4. We will merge this PR to develop and build on top of this PR incrementally next week for final polishes

1. Resolved two TODOs (sample notifications, govt masthhead height) - attempted to resolve the exit confirmation by [setting a form default value and checking if form state isDirty](https://react-hook-form.com/docs/useformstate), but doesn't seem to work on custom objects
2 & 4. I think FF would have been a good idea. Unfortunately I didn't do that to begin with and won't have time to do the reverts and introduce FF proper. Since we don't have any urgent releases planned and the revamp is timeboxed to 2 weeks, I think it should be ok to have the current state in develop, wdyt @PikkaPikkachu ? Alternative can be keeping all revamp work in a separate branch and merging to that branch for now over the next 2 weeks, so that urgent releases can still go into develop directly.

gweiying:
> EDIT: As discussed over slack, we are planning to write E2E tests before releasing, so it should be fine!
yep I've only disabled E2E tests passing for merge to develop - merging to production should still be e2e tested!

PR #302, opened 2023-09-26
feat: highlight placeholders

ameliatay:
## Context

1. Users have difficulty identifying where the placeholders are on the html template
2. To pre-empt users on the placeholders they need to fill up

Closes [this issue](https://www.notion.so/opengov/feat-Issuance-flow-select-template-Highlight-variables-e2e34db77741426b9197264d91e51f45?pvs=4)

## Approach

Before passing the HTML into the LetterViewer component, if PreviewLetter component receives a prop for isPlaceholdersHighlighted set to `true`, then the following replacements will happen:

- `{{` -> `<mark>{{`
- `}}` -> `}}</mark>`

This will highlight components in yellow

## Sample Screenshot
Screenshot 2023-09-26 at 10 19 48 AM
Screenshot 2023-09-26 at 10 24 13 AM


gweiying (approved):
lgtm!

PR #314, opened 2023-09-27
fix: reduce error on issue template button clicked

ameliatay:
## Context

Application crashes when templates are not retrieved yet but the dropdown (select a template, issuance flow) is rendering

## Approach

Add conditional to check if templates exists before rendering dropdown `SingleSelect` component


ameliatay:
> Hmm this might not be the best UI — we will get an issuance flow without any drop down indication, while it does prevent the crash from happening. Maybe let's think of a more user friendly solution, like letting them know that no templates exist or providing them with a default template?

Ahh understand. I think your second suggestion sounds great so they can at least check out the flow. I'll go ahead to implement this first as a quick fix while we wait for design decision on (1) How to deal with cases with no templates and (2) a test flow

ameliatay:
If user has no templates, add a sample template for them to view.

Assumes: Changes for access control will only be on the BE
Tested: In TypeORM, when find() can't locate any records, it returned `undefined`

Hence, if data is `undefined`, return `[sampleTemplate]` instead

ameliatay:
Awaiting design decision and it likely will not be using this solution. develop-v3.0 branch has also gone through some changes so effort to edit this branch and rebase VS creating a new branch once design is out is too high. Will be closing this PR.

PR #316, opened 2023-09-27
Chore/theme setup

ameliatay:
## Context

In preparation of changing Letter's theme, set up design system to make the change easier.

Closes [this issue](https://www.notion.so/opengov/chore-Set-up-themes-be28bc1a1f6d4b479c97b0440fb0dde9?pvs=4)

## Approach

- Set up mock generated themes for the moment (Need Jue Ling to generate her desired theme using Figma Tokens in the future)
- Removed hardcoded colors


frontend/src/app/App.tsx
```diff
@@ -4,7 +4,7 @@ import '@fontsource/ibm-plex-mono'
 import { ThemeProvider } from '@opengovsg/design-system-react'
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
 
-import { theme } from '~/theme'
+import { theme } from '~/theme/theme'
```
gweiying, line 7: tiny bug here, i believe the file reference needs to be changed to `'~/theme'`!

frontend/src/features/dashboard/components/batchViewComponents/IssuedBatchViewLetterTable.tsx
```diff
@@ -40,33 +40,33 @@ export const IssuedBatchViewLetterTable = ({
...
             size="sm"
           >
             <Tr
               textStyle={'subhead-2'}
-              borderBottomWidth={'1.5px'}
+              borderBottomWidth={'2px'}
+              borderTopWidth={'2px'}
               borderColor={'grey.200'}
-              backgroundColor={'white'}
+              backgroundColor={'standard.white'}
```
gweiying, line 69: for my own learning, why are we using `'standard.white'` here instead of `'white'`, is there a difference?

frontend/src/features/issue/components/Stepper.tsx
```diff
@@ -38,7 +38,7 @@ const Step = ({ label, description, index }: StepProps) => {
         height="3px"
         color="red"
```
gweiying, line 39: i built this but forgot to remove this, was for debugging purporses! could you help remove line 39?

frontend/src/features/dashboard/components/batchViewComponents/IssuedBatchViewLetterTable.tsx
ameliatay: Was ensuring that all our colour references throughout the codebase use something already defined in our `colors.ts` file. Since Letters has yet to define our own palette, I took ref from palettes in other ogp products, where grey is defined as `grey` whereas white is defined as `standard.white`. Generally, it was a best practice in prev companies i was with, but after scanning github I think it's less enforced in OGP though some products like phonebook do this too^ (i.e. Not using white from chakra's palette, but using standard.white from our own palette)

ameliatay:
WIP - Removed the fontsource from main file because it should be imported from themes, but seems like it wasn't, working on it!

frontend/src/features/dashboard/components/batchViewComponents/IssuedBatchViewLetterTable.tsx
gweiying: Thanks for explaining!

gweiying (approved):
Overall colors looks good! Thanks for combing through the codebase so thoroughly

I think we should also look into base and med screen size support too, setting sizes for both base and med screen sizes [p1]. We can also look into hardcoded spacing (move away from hardcoded px or rem) too but that one is less urgent [p2]

ameliatay:
> Overall colors looks good! Thanks for combing through the codebase so thoroughly
> 
> I think we should also look into base and med screen size support too, setting sizes for both base and med screen sizes so that we can make sure that UX behavior on GSIB is good too [p1]. We can also look into hardcoded spacing (move away from hardcoded px or rem) too but that one is less urgent [p2]

Ok! Will leave that P1 task to the current pending task we have for GSIB and create a new one for P2

PR #318, opened 2023-09-29
chore: split issued batches table into seperate com

gweiying:
## Context
Minor clean up to split out the issued batches table into a separate component from the page view

No functional or visual changes here.


gweiying:
Closing as duplicate with #319

PR #324, opened 2023-10-02
chore: issue confirmation modal and download

disKeith:
## Context

As part the UX revamp, this PR updates the issuance confirmation modal content and visuals, and automatically downloads the completed csv for the manual notification issuance flow. 

Ref: [Figma](https://www.figma.com/file/KVFsGATLUfvPqKE2lY9niH/LetterSG?type=design&node-id=2847%3A101047&mode=dev)

## Screenshots
**SMS**
confirm_SMS

**Email**
confirm_Email

**Manual**
confirm_Manual

PR #326, opened 2023-10-03
chore: refactor bulk create endpoint

disKeith:
## Context

This PR refactors the `{{host}}/api/letters/bulks` endpoint to return the newly created batch instead of bulk letters, and to remove the frontend workaround for downloading batches on manual creation

Discussed in: https://github.com/opengovsg/letters/pull/324#discussion_r1343405070


backend/src/letters/letters.controller.ts
```diff
@@ -51,23 +53,20 @@ export class LettersController {
...
         user.id,
         letters,
         createNotificationDto,
       )
-    return mapLetterToGetBulkLetterDto(
-      bulkRequest.letterParamMaps,
-      letters,
-      bulkRequest.passwords,
-    )
+    return this.batchesService.findOne(batch.id)
```
gweiying, line 69: might want to leave a comment above why we chose to do this, could be good to refactor down the road for cleanliness

gweiying (approved):
lgtm!

PR #329, opened 2023-10-03
fix: new user padding

gweiying:
Removes px from NewUserCard


ameliatay (approved):
lgtm! thanks for the changee

PR #348, opened 2023-10-05
fix: bugs in bulk issuance flow [Part 1]

PikkaPikkachu:
Fixes the template scaling issue. The scaling we were doing was previously incorrect.

Also reduced the border radius to 10px for the letter preview.


frontend/src/features/issue/components/previews/PreviewTemplate.tsx
```diff
@@ -64,20 +64,19 @@ export const PreviewTemplate = ({
         <Box
           backgroundColor="white"
           margin="60px"
-          marginBottom="0px"
           marginTop={'60px'}
-          borderTopRadius={'20px'}
+          borderTopRadius={'10px'}
```
gweiying, line 68: nit from ben was that bottom of letter should also have radius and blue background, would that be easy to achieve or no? if radius but no background that would be pretty difficult too

gweiying: let's handle that outside this PR anyway!

gweiying (approved):
seems good

PR #382, opened 2023-10-20
fix: template preview index issue when new templates are created

gweiying:
## Problem
Previously, because of the way the selected template was being managed, the selected template would not update in the preview when a new template was created.

Changes:
- Moves `TemplatePreviewModal` to sit inside `TemplatesBody` of `TemplateCard` since `TemplatePreviewModal` could be previewing any template, not just the one tied to the `TemplateCard`
- Minor cleanups and removing unnecessary `useForm` method

## Before & After Screenshots

**BEFORE**:
Screen Recording 2023-10-20 at 5 56 15 PM

**AFTER**:

Screen Recording 2023-10-20 at 5 52 29 PM


frontend/src/features/template-library/TemplatePreviewModal.tsx
```diff
@@ -15,39 +12,24 @@ import {
 interface TemplatePreviewModalProps {
   isOpen: boolean
   onClose: () => void
-  template: GetTemplateDto
-  index: number
+  templates: GetTemplateDto[]
+  initialIndex: number | null
```
ameliatay, line 16: just for learning!! why does this need to be `number | null`? if the only access point for the overlay/preview is on TemplateCard click, is it right to say there would always be an index?

ameliatay (approved):
lgtm! thanks for refactoring my initial implementation, looked through to see what i could do better too 💯

frontend/src/features/template-library/TemplatePreviewModal.tsx
gweiying: Good point, I refactored it so that `initialIndex` is always non-null

